PR #1, opened 2015-01-27
bludgeon: use system-installed wp-cli.phar

chriskuehl:
Puppet installs wp-cli system-wide on login servers, so we can just use that.

See also ocf/puppet@1d9eae87d27283aabbf3959f6b0212fb43096b58 and 596bf1fb33f6cec0914bc670980841737e0cc5e9


nattofriends:
nice

PR #3, opened 2015-02-11
bludgeon: don't run as root, read pass from config

chriskuehl:
Currently running bludgeon as root works great but will create files owned by root in the user's web directory, so you have to chown it afterwards. Instead, we can run it as the user and read their password out of wp-config.php.

Potentially users could run this too, but I doubt they will.

### DIFF
--- a/staff/web/bludgeon
+++ b/staff/web/bludgeon
@@ -2,45 +2,33 @@
 """
 Force the Disable Comments plugin onto an existing Wordpress install.
 
-This needs to be run as root; try
+This should be run as the appropriate user. Try:
 
-    sudo `which bludgeon.py` /path/to/user/wp
+    sudo -u someuser bludgeon /path/to/user/wp
 
-This also needs a MySQL root password; you can provide it on the
-command line or use a standard MySQL configuration file
-(default ~/.my.cnf)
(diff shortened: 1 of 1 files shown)


staff/web/bludgeon
```diff
@@ -123,37 +111,32 @@ class Wordpress(object):
...
+    against a malicious WordPress install, but so is wp-cli."""
 
-    args = parser.parse_args()
+    config_path = join(path, 'wp-config.php')
+    if not os.path.isfile(config_path):
+        raise Exception("Can't find WordPress installation.")
 
-    if os.geteuid() != 0:
-        parser.error("This utility should be run as root.")
+    code = "require('" + config_path + "'); echo DB_PASSWORD;"
```
chriskuehl: this is potentially dangerous but it's probably ok

chriskuehl:
@nattofriends is this how you were doing it? `php -F filename -E something` seemed promising, but seemed like it was more for running a filter on a line (awk-style). This has potentially escaping problems but I doubt it will ever matter.

nattofriends:
Yea that's how I was doing it I think (with -E at least), your proposal works too

PR #4, opened 2015-10-29
Rewrite signat with ocflib

matthew-mcallister:
Waiting on a pull request in ocflib and needs python-tabulate to be installed, but it works nicely in a virtualenv.

### DIFF
--- a/staff/acct/signat
+++ b/staff/acct/signat
@@ -1,104 +1,177 @@
-#!/usr/bin/env python
-"""Returns active student group information for a signatory.
-OCF username or CalNet UID is the first and only argument.
-
-Required packages: python-ldap, python-yaml
+#!/usr/bin/env python3
+"""Returns active student group information about a signatory or
+student group.
 """
-
-import ldap
+import argparse
 import sys
-from urllib import urlencode
-from urllib2 import urlopen, URLError
-from xml.etree import ElementTree as etree
-from yaml import safe_dump as dump
 
-service = 'SignatoriesActiveStudentGroups'
-url_base = 'https://studentservices.berkeley.edu/WebServices/StudentGroupServiceV2/Service.asmx'
-ldap_uri = 'ldaps://ldap.ocf.berkeley.edu'
-ldap_basedn = 'ou=People,dc=ocf,dc=berkeley,dc=edu'
-ldap_filter = 'uid'
-ldap_attribute = 'calnetUid'
+import ocflib.account.search as search
+import ocflib.constants as constants
+import ocflib.infra.ldap as ldap
+import ocflib.ucb.directory as directory
+from ocflib.ucb.groups import (
+    list_groups,
+    group_by_oid,
+    signatories_for_group,
+    groups_by_student_signat,
+)
+from ocflib.misc.shell import red
+from tabulate import tabulate
 
-def get_osl(id, service, key='UID'):
-    """Query web service for student group information in XML format"""
-    if not isinstance(id, int):
-        raise ValueError(("ID must be an integer: " + str(id)))
-    if not service.isalpha():
-        raise ValueError(("Service is not alphabetic: " + str(service)))
-    parameters = {key: id}
-    url = url_base + '/' + service + '?' + urlencode(parameters)
-    try:
-        response = urlopen(url)
-    except URLError as e:
-        raise Exception("Could not connect to student group web service: "
-                        + str(e) + " at " + url)
-    return etree.parse(response)
 
-def parse_osl(root):
-    """Assemble Python dictionaries of groups from XML document"""
-    if root.findtext('Succeeded') == 'false':
-        try:
-            error_reason = root.findtext('Reason')
-        except:
-            error_reason = 'unknown reason'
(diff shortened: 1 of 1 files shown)


staff/acct/signat
```diff
@@ -1,104 +1,173 @@
...
-    else:
-        raise ValueError(("Account not found in LDAP: " + username))
+
+from tabulate import tabulate
+
+import ocflib.account.search as search
+import ocflib.constants as constants
+import ocflib.infra.ldap as ldap
+import ocflib.ucb.directory as directory
+from ocflib.ucb.groups import *
```
chriskuehl: wildcard imports are yucky, can you replace this?

staff/acct/signat
```diff
@@ -1,104 +1,173 @@
...
-        raise ValueError(("Invalid DN: " + ldap_basedn))
-    if entry:
-        try:
-            return int(entry[0][1][ldap_attribute][0])
-        except KeyError:
-            raise KeyError(("Account does not have a CalNet UID: " + username))
-    else:
-        raise ValueError(("Account not found in LDAP: " + username))
+
+from tabulate import tabulate
```
chriskuehl: nit: generally python imports are organized into three blocks, in this order:
1. stdlib imports
2. 3rd-party library imports
3. modules from this library/application
(with a blank line in between each block of imports)

so tabulate would be grouped alphabetically in the same block as the ocflib imports

but I wouldn't worry about it for now. It's a pain to do manually and pre-commit will do it for us once I get around to patching it for our use case.

staff/acct/signat
```diff
@@ -1,104 +1,173 @@
...
+
+    for (oid, attrs) in groups.items():
+        signatories = signatories_for_group(oid)
+        if signatories:
+            found = True
+            show_signatories_for_group(oid)
+            print()
+
+    if not found:
+        print(red("No groups had signatories."))
```
chriskuehl: I think you can probably rewrite this as:

```python
for oid, attrs in groups.items():
    signatories = signatories_for_group(oid)
    if signatories:
        show_signatories_for_group(oid)
        break
else:
    print(red("No groups had signatories."))
```

Also, maybe pass the signatories in to that function so we don't look them up twice?

staff/acct/signat
matthew-mcallister: Yeah, can't rely on pre-commit to fix up my imports on this repo.

staff/acct/signat
```diff
@@ -1,104 +1,173 @@
...
+
+    print("Searching for signatories...")
+
+    found = False
+
+    for uid in uids:
+        groups = groups_by_student_signat(uid)
+        if groups:
+            found = True
+            show_groups_by_student_signat(uid, groups)
```
chriskuehl: second arg should be a keyword arg?

staff/acct/signat
```diff
@@ -1,104 +1,173 @@
...
+            print()
+
+    if not found:
+        print(red("No signatories found."))
```
chriskuehl: you can rewrite the for loop similar to above if you care to

staff/acct/signat
matthew-mcallister: I was under the impression things got organized like:
1. stdlib
2. 3rd-party external to the developer
3. other libs from the same team/codehouse
4. this codebase

The convention you suggest makes enough sense though.

chriskuehl:
lgtm, you'll want to use apt-dater to ensure ocflib gets updated first, and you'll want to add python3-tabulate to puppet

note that python3-tabulate is not available on wheezy so for the time being you'll need to put it in an if block. i think it's fine if signat doesn't work on wheezy.

chriskuehl:
this is a pretty great change, think you can talk about this at the next BoD?

staff/acct/signat
matthew-mcallister: I don't think that will result in the same behavior. Won't that terminate when the first group with signatories is found?

And the latter is just a typo, thanks for catching.

staff/acct/signat
chriskuehl: I'm not familiar with those blocks, but in any case, I probably wouldn't worry about it personally. pre-commit will fix it to some standard style when I eventually get it working here.

(The reason we can't use pre-commit in utils yet is that it currently targets files based on extension, and almost nothing in utils has an extension. But I'm [working on a branch](https://github.com/pre-commit/pre-commit/compare/master...chriskuehl:hooks-by-type) that fixes that which I'll be deploying soon to OCF as a testing ground.)

staff/acct/signat
chriskuehl: you're right, I misread your code. looks fine as-is in that case.

staff/acct/signat
matthew-mcallister: Sure, why not.

matthew-mcallister:
I'll go over this at BoD. `approve` is basically a two-step process now.

matthew-mcallister:
Amended and ready to merge once Puppet is updated for deployment. In the meantime, open for comments on interface, documentation, speed, etc.

PR #6, opened 2015-11-01
Add pre commit (types fork)

chriskuehl:
See https://github.com/ocf/ocfweb/pull/75 for full details about the pre-commit branch.

Previously, it didn't really make sense to run pre-commit over utils because nearly all the scripts are extensionless, and it wasn't able to target them.

My branch, which I'm hoping to upstream soon, allows for determining file types by shebang, among other things. So it does a pretty good job on this repo.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -22,7 +22,7 @@ var/
 *.egg
 
 # PyInstaller
-#  Usually these files are written by a python script from a template 
+#  Usually these files are written by a python script from a template
 #  before PyInstaller builds the exe, so as to inject date/other infos into it.
 *.manifest
 *.spec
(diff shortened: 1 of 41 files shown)


matthew-mcallister:
Pre-commit nits: It doesn't seem to do anything for bash scripts. We also now have a de facto all single-quote convention in Python.

Otherwise, it's clearly quite functional. Unless you want to take @nattofriends' advice in https://github.com/ocf/ocfweb/pull/75, seems OK to ship.

chriskuehl:
@matthew-mcallister regarding shell scripts, the new branch will tag them properly:

    ckuehl@supernova:~/ocf-proj/utils$ pre-commit identify update-symlinks
    {'file', 'executable', 'text', 'shell'}


...which means hooks that target any of those tags will run. For example, tab-replacing or whitespace-trimming hooks do run.

The main issue is just there aren't a lot of high-quality linters for shell scripts that I know of. [This](https://github.com/koalaman/shellcheck) looks quite nice but sadly it's written entirely in Haskell which isn't supported by pre-commit (currently supports python, nodejs, ruby, pcre, and system executables).

PR #10, opened 2015-11-28
Convert paper command to use bash instead of perl

jvperrin:
The new command is functionally equivalent to the old paper script but with a few small changes:

1. The username and the -v flag passed to the script can now be interchanged, so the order does not matter.
2. If the script cannot find a user or the username is formatted badly (not all letters), then the new script will exit with an error code (1) instead of the perl script's exit code of 0.

Otherwise, the main changes apart from the language conversion are that some unused variables have been removed and the history comments in the old perl script have been removed too. The history will stick around in git, so that should be fine.

### DIFF
--- a/printing/paper
+++ b/printing/paper
@@ -1,96 +1,75 @@
-#!/usr/bin/perl
+#!/bin/bash
 
-# pykota not installed on tsunami, display a sane message in the interim
-#print "Printing quota records are temporarily unavailable";
-#exit;
+PKUSERS_SEMESTER='/opt/ocf/packages/pykota/pkusers --config /etc/pykota/pykota.semester/'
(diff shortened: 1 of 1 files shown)


chriskuehl:
no more perl? :)



ship it

printing/paper
```diff
@@ -1,96 +1,75 @@
...
+if [[ $1 && $1 == "-v" ]]; then
+  verbose=1
+  user=$2
+elif [[ $2 && $2 == "-v" ]]; then
+  verbose=1
+  user=$1
+else
+  verbose=0
+  user=$1
+fi
```
chriskuehl: heh, parsing command line arguments is usually my threshold for dropping bash and using python instead. this looks fine though.

jvperrin: Yeah, there's weren't many to deal with, so that was good.

PR #11, opened 2015-12-04
Notify sorried user by email

daradib:
Only sends an email if we have one on file.
Use email footer defined in ocflib constants.

### DIFF
--- a/staff/acct/sorry/sorry
+++ b/staff/acct/sorry/sorry
@@ -104,6 +104,27 @@ ls -ld $httpdir
 #logging
 echo `/bin/date` - $rootstaffer $sorriedUser >> /opt/acct/sorry.log
 
+# Notify user by email if email address is available.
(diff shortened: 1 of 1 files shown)


chriskuehl:
looks good to me
I'll assume you tested it

thanks and go to bed

daradib:
So, what's the roadmap on moving sorry/unsorry to ocflib :)

chriskuehl:
Want to do it but currently don't have a way to do ldap modifies in python3. 

The ldap3 library has gssapi support but I've never been able to get it working. 

For create we shell out to ldapadd currently. 

If you can get gssapi working you would be my hero.

chriskuehl:
https://github.com/ocf/ocflib/blob/master/ocflib/infra/ldap.py#L55

daradib:
For now it looks like we'll have to live with shelling out. I remember your struggles.

PR #16, opened 2016-02-29
Kill /opt/ocf and slightly update makemysql

matthew-mcallister:
Sorry this diff is kind of noisy. I made some changes to follow ocflib conventions and wound up making several unimportant line changes as well.

### DIFF
--- a/makeservices/makemysql-real
+++ b/makeservices/makemysql-real
@@ -1,153 +1,74 @@
-#!/usr/bin/python
-#
-# This script creates a MySQL database with the same name as the username
-# of the user who runs this program. To prevent the user from taking control
-# of this program and e.g., causing the program to dump core and reveal the
-# MySQL root password, this program should be run under another account,
-# preferably that of the MySQL user.
-#
-# The password set for this new account is randomly generated; it is not
-# user-selectable. This is to prevent compromised MySQL-backed web apps from
-# also compromising the shell account password. Note that the MySQL server
-# should also be patched to prevent the user from later changing the password.
-#
-# The MySQL root password and the hostname of the MySQL server are read from
-# a config file; the path to it is stored in the CONF_FILE global. This file
-# must be readable only by the setuid user! The Python ConfigParser module
-# is used to parse the config file, so the format is basically the Windows
-# INI format.
-#
-from __future__ import print_function
+#!/usr/bin/env python3
+"""Creates a MySQL database with the same name as the username of the user who
+runs this program.
 
+To prevent the user from taking control of this program and e.g., causing the
+program to dump core and reveal the MySQL root password, this program should be
+run under another account, preferably that of the MySQL user.
+
+The password set for this new account is randomly generated; it is not
+user-selectable. This is to prevent compromised MySQL-backed web apps from
+also compromising the shell account password. Note that the MySQL server
+should also be patched to prevent the user from later changing the password.
+
+The MySQL root password and the hostname of the MySQL server are read from
+a config file; the path to it is stored in the CONF_FILE global. This file
+must be readable only by the setuid user! The Python configparser module
+is used to parse the config file, so the format is basically the Windows
+INI format.
+"""
 import os
+import random
+import string
 import sys
+from configparser import ConfigParser
+from textwrap import dedent
 
-import ConfigParser
 import MySQLdb
-
-#
-# Global variable definitions.
-#
-
-RANDOM_PW_CMD = 'pwgen -n1 -y -s 20'
+from ocflib.misc.mail import send_problem_report
 
 
-WELCOME_TEMPLATE = '/opt/ocf/etc/makemysql.template'
-WELCOME_SUBJECT = 'MySQL account information'
-WELCOME_SENDER = 'staff@OCF.Berkeley.EDU'
+CREATE_QUERY = 'CREATE DATABASE {}'
+GRANT_QUERY = "GRANT ALL PRIVILEGES ON {username}.* TO '{username}'@'%' IDENTIFIED BY '{userpass}'"
 
-MAIL_USER_CMD = "/usr/bin/sed -e 's/FOO/%%s/' < %s | \
-        /usr/bin/mailx -s '%s' -r '%s' %%s" % \
-                (WELCOME_TEMPLATE, WELCOME_SUBJECT, WELCOME_SENDER)
+CONF_FILE = '/opt/share/makeservices/makemysql.conf'
+MYSQL_HOST = 'mysql.ocf.berkeley.edu'
 
-
-GRANT_QUERY = "GRANT ALL PRIVILEGES ON %s.* TO %s@'%%' IDENTIFIED BY '%s'"
-
-CREATE_QUERY = 'CREATE DATABASE %s'
-
-CONF_FILE = '/opt/ocf/etc/makemysql.conf'
-MYSQL_HOST = 'maelstrom.int.ocf.berkeley.edu'
-MYSQL_ROOT_PW = None  # No default; this must be read from the config file.
-
-#
-# For debugging purposes.
-#
-
-DEBUG_MODE = False
-DEBUG_EMAIL_ADDR = 'wheel@OCF.Berkeley.EDU'
-
-#
-# Function and class definitions.
-#
-
-
-def send_error_email(username, exc_info):
-    import smtplib
-    import string
-    import traceback
-    from_addr = 'root@OCF.Berkeley.EDU'
-    if DEBUG_MODE:
-        to_addr = DEBUG_EMAIL_ADDR
-    else:
-        to_addr = 'root@OCF.Berkeley.EDU'
-    msg = 'From: makemysql at OCF <%s>\n' % from_addr
-    msg += 'To: MySQL <%s>\n' % to_addr
-    msg += "Subject: 'makemysql' unhandled exception\n"
-    msg += '\n'
-    msg += "Username: '%s'\n\n" % username
-    msg += 'An unhandled exception has occurred:\n\n'
-    msg += string.join(traceback.format_exception(*exc_info), '')
-    msg += '\n<EOT>\n'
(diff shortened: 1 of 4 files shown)


makeservices/makemysql-real
```diff
@@ -1,153 +1,110 @@
...
-
-
-GRANT_QUERY = "GRANT ALL PRIVILEGES ON %s.* TO %s@'%%' IDENTIFIED BY '%s'"
-
-CREATE_QUERY = 'CREATE DATABASE %s'
-
-CONF_FILE = '/opt/ocf/etc/makemysql.conf'
-MYSQL_HOST = 'maelstrom.int.ocf.berkeley.edu'
+CONF_FILE = '/opt/share/makeservices/makemysql.conf'
+MYSQL_HOST = 'maelstrom.ocf.berkeley.edu'
```
chriskuehl: prefer the service CNAMEs, e.g. mysql.ocf.berkeley.edu or just mysql

makeservices/makemysql-real
```diff
@@ -1,153 +1,110 @@
...
-# Read in the configuration file and set the appropriate globals.
+    subject = "'makemysql' unhandled exception"
+    body = dedent(
+        """\
+        Username: '{username}'
+        An unhandled exception has occurred:
+        {exc_info}
+        """
+    ).format(username=username, traceback=exc_info)
+    send_mail(MAIL_ROOT, subject, body, sender=MAIL_ROOT)
```
chriskuehl: can you just use ocflib's send_problem_report?

makeservices/makemysql-real
```diff
@@ -155,85 +112,88 @@ def main():
...
-            query = GRANT_QUERY % (username, username, userpass)
+            query = GRANT_QUERY.format(username=username, userpass=userpass)
             connection.query(query)
             connection.store_result()
             # Result should be "Query OK, 0 rows affected",
             # but we'll assume no exception means success.
         except:
-            print('>>> Error: Failed to grant database privileges.')
-            sys.exit(1)
+            print('>>> Error: Failed to grant database privileges.\n')
```
chriskuehl: why do you need \n here?

makeservices/makemysql-real
```diff
@@ -155,85 +112,88 @@ def main():
...
-main()
+        send_error_email(username, exc_info())
+        print(dedent(
+            """\
+            {}
+
+            A fatal error was detected during program execution.
+            OCF staff have been notified of the problem.
+            """
+        ).format(exc_info()))
```
chriskuehl: i don't think we want to print tracebacks to users, it's possible that the password might get leaked in them somehow. we need to be really careful with this script.

staff/acct/sorry/sorry
```diff
@@ -54,7 +54,7 @@ else
     export KRB5CCNAME
 fi
 
-sorryshell=/opt/ocf/bin/sorried
+sorryshell=/opt/share/utils/bin/sorried
```
chriskuehl: this probably needs to be updated across a lot of repos and also in ldap

makeservices/makemysql-real
```diff
@@ -155,85 +112,88 @@ def main():
...
             database_already_exists(username)
         except:
-            # The database already exists, so skip the creation step.
-            print('no.')
+            print('no.\n')
             db_create = True
 
         # Add or update user database privileges.
         try:
             userpass = os.popen(RANDOM_PW_CMD).read().strip()
```
chriskuehl: can we just generate the password in python rather than subprocess? the simpler this script the better

```python
import string, random
random_pw =''.join(random.SystemRandom().choice(string.ascii_letters + string.digits) for _ in range(20))
```

otherwise we should at least switch to subprocess

chriskuehl:
delaying shipit on this until i have time to look at makemysql closely, hope that's okay. it's a pretty important script and easy to make mistakes that leak the password.

makeservices/makemysql-real
matthew-mcallister: Got it

makeservices/makemysql-real
matthew-mcallister: You're right, I thought that was a part of ocfweb and not ocflib, I'm silly.

makeservices/makemysql-real
matthew-mcallister: It's not needed. I think I had some other print statements at some point and wanted the extra newline. I'll undo such unnecessary line changes.

makeservices/makemysql-real
matthew-mcallister: I'll change that. It was in the original version FTR.

staff/acct/sorry/sorry
matthew-mcallister: I've done it here and in puppet; probably should go over every repo just to be sure. I almost forgot about LDAP.

makeservices/makemysql-real
matthew-mcallister: Yeah, that was something I meant to change. I wasn't sure of how to do it properly though; I also meant to ask that.

matthew-mcallister:
Updated diff.

makeservices/makemysql-real
```diff
@@ -1,153 +1,78 @@
...
-# The MySQL root password and the hostname of the MySQL server are read from
-# a config file; the path to it is stored in the CONF_FILE global. This file
-# must be readable only by the setuid user! The Python ConfigParser module
-# is used to parse the config file, so the format is basically the Windows
-# INI format.
-#
-from __future__ import print_function
-
+#!/usr/bin/python3
+""" Creates a MySQL database with the same name as the username of the user who
```
chriskuehl: nit: `#!/usr/bin/env python` for shebang
nit: no space at the start of docstrings.

makeservices/makemysql-real
```diff
@@ -1,153 +1,78 @@
...
-
-GRANT_QUERY = "GRANT ALL PRIVILEGES ON %s.* TO %s@'%%' IDENTIFIED BY '%s'"
-
-CREATE_QUERY = 'CREATE DATABASE %s'
-
-CONF_FILE = '/opt/ocf/etc/makemysql.conf'
-MYSQL_HOST = 'maelstrom.int.ocf.berkeley.edu'
+CONF_FILE = '/opt/share/makeservices/makemysql.conf'
+MYSQL_HOST = 'mysql.ocf.berkeley.edu'
 MYSQL_ROOT_PW = None  # No default; this must be read from the config file.
```
chriskuehl: i don't think this global is used anymore?

makeservices/makemysql-real
```diff
@@ -1,153 +1,78 @@
...
+        see it when it appears.
+
+        If you are ready to continue, type 'yes'.
+        Typing anything other than 'yes' will abort this script.
+
+        Continue?
+        """
+    ))
+    inp = sys.stdin.readline().strip()
+    return inp == 'yes'
```
chriskuehl: nit: normally prefer to use `input('Continue? ') == 'yes'` because then they can use readline keyboard shortcuts. nbd here.

makeservices/makemysql-real
matthew-mcallister: I misunderstood what the purpose of /usr/bin/env was.

makeservices/makemysql-real
matthew-mcallister: Good idea.

makeservices/makemysql-real
```diff
@@ -155,85 +80,105 @@ def main():
...
 
         # Connect to the MySQL server.
         try:
             print('>>> Connecting to MySQL database server...')
-            connection = MySQLdb.connect(host=MYSQL_HOST,
+            connection = MySQLdb.connect(host=mysql_host,
                                          user='root',
-                                         passwd=MYSQL_ROOT_PW)
-        except:
+                                         passwd=mysql_root_pw)
```
chriskuehl: TODO: we should use pymysql for this eventually

makeservices/makemysql-real
```diff
@@ -155,85 +80,105 @@ def main():
...
+        except Exception as ex:
             print('>>> Error: Failed to connect to MySQL server.')
-            sys.exit(1)
+            raise ex
```
chriskuehl: nit: you can just use `raise` here if you feel like it

makeservices/makemysql-real
```diff
@@ -155,85 +80,105 @@ def main():
...
-#
-# Entry point.
-#
-main()
+            print(dedent("""
+            Your MySQL database has been created.
+
+            For instructions on accessing and using your database, please visit
+
+              http://ocf.berkeley.edu/docs/services/mysql/
```
chriskuehl: nit: `https://www.`

makeservices/makemysql-real
chriskuehl: it doesn't help too much here, but it's useful for python scripts which we often want to test in virtualenvs (e.g. imagine we wanted to test against a new version of pymysql or something)

chriskuehl:
mostly just nits above, i say shipit (but make sure to update sorried users' shells in LDAP at the same time)

makeservices/makemysql-real
matthew-mcallister: The exception gets used to send a problem report.

chriskuehl: `raise` by itself will re-raise the same exception, so it will still work.

you can actually just write in this case:

```python:
except:
    print('>>> Error: Failed to connect to MySQL server.')
    raise
```

matthew-mcallister:
OK, should be the last diff update.

makeservices/makemysql-real
```diff
@@ -1,153 +1,76 @@
...
+        web applications are exploited.
 
+        The randomly-generated password will be displayed on-screen.
+        Please make sure you are in an environment where nobody else will
+        see it when it appears.
+
+        If you are ready to continue, type 'yes'.
+        Typing anything other than 'yes' will abort this script.
+
+        Continue?
```
chriskuehl: remove this continue now?

PR #18, opened 2016-09-05
Rewrite makehttp in python

matthew-mcallister:
I'm pretty surprised this wasn't already done. Maybe I should have assigned it to someone else.

### DIFF
--- a/makeservices/makehttp
+++ b/makeservices/makehttp
@@ -10,13 +10,11 @@ symlinks or renaming any file or directory with the same path.
 import datetime
 import os
 import os.path
-import sys
+from argparse import ArgumentParser
 
-USER_SERVICES = '/services/http/users/{user[0]}/{user}'
-
-
-def usage(args):
-    return 'Usage: ' + args[0]
+from ocflib.account.utils import home_dir
+from ocflib.account.utils import web_dir
+from ocflib.misc.whoami import current_user
 
 
 def rectify_link(path, webdir):
@@ -42,12 +40,12 @@ def rectify_link(path, webdir):
 
 
 def main():
-    if len(sys.argv) > 1:
-        print(usage(sys.argv))
-        return
+    parser = ArgumentParser(description=__doc__)
+    parser.parse_args()
 
-    path = os.path.expanduser('~/public_html')
-    webdir = USER_SERVICES.format(user=os.getlogin())
+    user = current_user()
+    path = home_dir(user) + '/public_html'
+    webdir = web_dir(user)
 
     rectify_link(path, webdir)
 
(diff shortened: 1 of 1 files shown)


makeservices/makehttp
```diff
@@ -1,53 +1,58 @@
...
+idempotent: if the symlink exists and points to the correct path, makehttp
+takes no action; otherwise, it creates the symlink after removing broken
+symlinks or renaming any file or directory with the same path.
+"""
+import datetime
+import os
+import os.path
+import sys
+
+USER_SERVICES = '/services/http/users/{user[0]}/{user}'
```
chriskuehl: you probably want `ocflib.account.utils.web_dir`

makeservices/makehttp
```diff
@@ -1,53 +1,58 @@
...
+    os.chmod(path, 0o755)
+
+
+def main():
+    if len(sys.argv) > 1:
+        print(usage(sys.argv))
+        return
+
+    path = os.path.expanduser('~/public_html')
+    webdir = USER_SERVICES.format(user=os.getlogin())
```
chriskuehl: `os.getlogin` is a quirky function that doesn't really do what people usually want it to:

```
ckuehl@supernova:~/ocf-proj/ocflib$ python -c 'import os; print os.getlogin()'     
ckuehl
ckuehl@supernova:~/ocf-proj/ocflib$ sudo python -c 'import os; print os.getlogin()'
ckuehl
ckuehl@supernova:~/ocf-proj/ocflib$ sudo -u ggroup python -c 'import os; print os.getlogin()'
ckuehl
```

This is doubly quirky because the expanduser will still expand to the "correct" place when using sudo (e.g. "/home/g/ggroup", and so the symlink is created in their home directory but pointing to your web directory).

I'd suggest using `ocflib.misc.whoami.current_user` or just `getpass.getuser` directly.

chriskuehl:
otherwise lgtm!

makeservices/makehttp
```diff
@@ -1,53 +1,58 @@
...
+import datetime
+import os
+import os.path
+import sys
+
+USER_SERVICES = '/services/http/users/{user[0]}/{user}'
+
+
+def usage(args):
+    return 'Usage: ' + args[0]
```
chriskuehl: Might be cool to print `__doc__` at the end so the useful docstring you wrote gets printed as well.

matthew-mcallister: I actually thought about that but didn't feel like adding a "-h" option

matthew-mcallister:
Nice, the return of the shipit gifs

PR #20, opened 2016-10-21
python2 -> python3 migration

douglaslwong:
@matthew-mcallister merging into master for testing

### DIFF
--- a/net/pyrc
+++ b/net/pyrc
@@ -2,7 +2,6 @@
 # pyrc - Wrapper to make it easier for ocf users to use irc
 # Author: Nader Morshed
 # Version: 0.0.8
-# Requires python >= 2.4 (2.6?)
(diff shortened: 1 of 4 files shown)


chriskuehl:
seems okay otherwise, i didn't test them or read too thoroughly though

matthew-mcallister:
Check out `pyrc.find_emu_sessions`. We don't need to check if the python version is >= 3. If you wanted, you could additionally change it to use `subprocess.check_output` instead of `Popen`.

Also, get rid of the comment that says "Requires python >= 2.6".

douglaslwong:
oops should've closed this PR last night, thanks for the feedbacks

douglaslwong:
changes finalized and tested, @matthew-mcallister or @chriskuehl?

PR #23, opened 2016-11-16
Fix compatibility with ldap3 version 2

matthew-mcallister:
See https://github.com/ocf/ocflib/pull/38.

### DIFF
--- a/staff/sys/ldap-lint
+++ b/staff/sys/ldap-lint
@@ -66,12 +66,12 @@ def main():
         )
         for attrs in map(itemgetter('attributes'), c.response):
             cn = attrs['cn'][0]
-            type_ = attrs['type'][0]
+            type_ = attrs['type']
 
             if type_ not in RECOGNIZED_TYPES:
                 complain(cn, 'has unknown type ' + type_)
 
-            if 'macAddress' in attrs:
+            if attrs['macAddress']:
                 mac_addr = attrs['macAddress'][0].lower()
 
                 if type_ != 'desktop':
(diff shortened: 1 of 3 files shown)


staff/sys/ldap-lint
```diff
@@ -66,12 +66,12 @@ def main():
...
         for attrs in map(itemgetter('attributes'), c.response):
             cn = attrs['cn'][0]
-            type_ = attrs['type'][0]
+            type_ = attrs['type']
 
             if type_ not in RECOGNIZED_TYPES:
                 complain(cn, 'has unknown type ' + type_)
 
-            if 'macAddress' in attrs:
+            if attrs.get('macaddress', []):
```
chriskuehl: this seems wrong? what's going on here?

matthew-mcallister: The capitalization is a typo, thanks for spotting. I tested it before refactoring and making the typo, though.

LDAP entries can have multiple `macAddress` attributes, so `attrs['macAddress']` is supposed to be a list. When an entry doesn't have a `macAddress` entry, `attrs['macAddress']` is an empty list. In the old version, `attrs` simply didn't have the key `'macAddress'`.

`ldap-lint` should probably be updated to complain if multiple `macAddress` attributes are defined. (Unless the schema supports specifying only zero or one of an attribute?)

matthew-mcallister: `attrs` always contains `'macAddress'` now, so I've updated the code to reflect that.

chriskuehl: we should probably fix the schema so that we can have only one macAddress

jvperrin: `ldap-lint` will have to be updated too, since it uses ocflib and there are breaking changes with the upgrade.

jvperrin: Lol never mind, apparently I can't read

PR #24, opened 2016-11-18
basic weather script

Baisang:
could also just use subprocess.call with curl instead of faking the user agent, not sure what is best approach


chriskuehl:
why is this under desktop?

chriskuehl:
otherwise lgtm

Baisang:
no particular reason, didn't know where would be a better place to put

desktop/weather
```diff
@@ -0,0 +1,32 @@
...
+    # otherwise we get html in our response
+    headers = {'User-Agent': 'curl'}
+    res = requests.get(WTTR_SERVER + location, headers=headers)
+    if res.status_code == 200 and res.text != WTTR_ERROR:
+        print(res.text)
+    else:
+        print('location not found')
+
+def main(args):
+    parser = optparse.OptionParser()
```
chriskuehl: argparse > optparse

chriskuehl:
okay, i guess i don't have a better idea besides making a "lol" directory or something

jvperrin:
Maybe put it in the `net` directory? This doesn't seem like it's desktop-specific, whereas `pdf-open` (the only current desktop script) is definitely desktop-specific.

Baisang:
sounds good to me

matthew-mcallister:
I am disappointed this does not work in tmux.

matthew-mcallister:
Also you can decide weather you want to squash those commits.

chriskuehl:
please don't start a pun thread
i can't take any more

chriskuehl:
i'm deleting any future comments containing puns

douglaslwong:
wtf is this

gnowxilef:
why does it not work in tmux?

chriskuehl:
wfm, provide details of failure?

matthew-mcallister:
It doesn't work for me in tmux either:

screenshot - 12022016 - 11 13 05 pm

PR #25, opened 2016-11-22
Use signat in check script

daradib:
Also fix indentation and move warning checks together.

### DIFF
--- a/acct/check
+++ b/acct/check
@@ -52,25 +52,25 @@ else
   exit 2
 fi
 
-# check if sorried
-shell=$(ldapsearch -x -LLL uid=$user loginshell | grep ^loginShell | cut -d' ' -f2)
-if [ $shell = /opt/share/utils/bin/sorried ]; then
-  echo "WARNING: Account is disabled" >&2
-fi
-
 # account creation time
 creation_time=$(ldapsearch -x -LLL uid=$user creationTime | grep ^creationTime | cut -d' ' -f2)
 if [ -n "$creation_time" ]; then
     echo "Created on: $(python3 -c "import dateutil.parser; print(dateutil.parser.parse('${creation_time}').strftime('%Y-%m-%d'))")"
 fi
 
+# check if sorried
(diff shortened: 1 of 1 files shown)


acct/check
```diff
@@ -52,25 +52,25 @@ else
...
+shell=$(ldapsearch -x -LLL uid=$user loginshell | grep ^loginShell | cut -d' ' -f2)
+if [ $shell = /opt/share/utils/bin/sorried ]; then
+  echo "WARNING: Account is disabled" >&2
+fi
+
 # check if Kerberos principal exists
 kerberos=$(echo TESTIFEXIST | kinit --password-file=STDIN $user 2>&1)
 if [ -n "$kerberos" ]; then
-  if [ -n "$(echo "$kerberos" | grep unknown$)" ]; then
+  if echo "$kerberos" | grep -q 'unknown$'; then
```
chriskuehl: can we just make this bash lol

daradib: feature creep ;)

acct/check
```diff
@@ -96,30 +96,24 @@ fi
...
-   echo "Member of group(s): $groups"
+  echo "Member of group(s): $groups"
 fi
 
-# check if CalLink OID number exists in LDAP
-callinkOid=$(ldapsearch -x -LLL uid=$user callinkOid | grep -i ^callinkOid | cut -d' ' -f2)
-if [ -n "$callinkOid" ]; then
-   echo "CalLink OID number: $callinkOid"
-   if [ "$callinkOid" != 0 ]; then
-       echo "https://ocf.io/callinkapi?status=&category=&type=&name=&OrganizationID=$callinkOid"
```
daradib: The URL? We can add it back but I think it belongs in `signat` instead.

chriskuehl: i don't understand why it was deleted?

tahabi:
I did this in #45 and neglected to check that this PR actually does the same thing, sorry about that :(

PR #27, opened 2016-12-12
stdin2syslog: use only one logger for stretch+

kpengboy:
`logger` can read input from stdin, but the functionality is broken
on jessie. For jessie, use our current behavior; for stretch and
above, just `exec logger`.

### DIFF
--- a/staff/sys/stdin2syslog
+++ b/staff/sys/stdin2syslog
@@ -1,49 +1,34 @@
-#!/usr/bin/env python3
-"""Write stdin logs to OCF's remote syslog server.
+#!/bin/bash
+set -euo pipefail
+
+if [ $# -lt 1 ] || [ "$1" = "-h" -o "$1" = "--help" ]; then
+    echo "Usage: $0 stream
+
+Write stdin logs to OCF's remote syslog server.
 
 You can use arbitrary stream names (without configuring them before); each
 stream will show up at syslog:/var/log/remote/<stream>.
-"""
-import argparse
-import re
-import subprocess
-import sys
+"
+    exit 1
+fi
 
+stream="$1"
 
-VALID_STREAM_NAME = re.compile('^[a-z0-9\-_]+$')
+grep -qE '^[a-z0-9_][-a-z0-9_]+$' <<< "$stream" || {
+    echo "$0: invalid stream name."
+    exit 2
+}
 
-
-class LoggingFailure(Exception):
-    pass
-
-
-def write_line(stream, line):
-    """Write a line to a stream.
-
-    This might make sense to move to ocflib some day.
-    """
-    if not VALID_STREAM_NAME.match(stream):
-        raise ValueError('Bad stream name: {}'.format(stream))
-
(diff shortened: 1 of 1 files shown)


kpengboy:
Another possibility is to try to backport `logger`. Since that comes as part of the wider `util-linux` source package, things could not be as straightforward...

staff/sys/stdin2syslog
```diff
@@ -1,49 +1,34 @@
-#!/usr/bin/env python3
-"""Write stdin logs to OCF's remote syslog server.
+#!/bin/bash
+set -euo pipefail
```
chriskuehl: -1

chriskuehl:
#28 is my attempt

staff/sys/stdin2syslog
```diff
@@ -1,49 +1,34 @@
...
+
+grep -qE '^[a-z0-9_][-a-z0-9_]+$' <<< "$stream" || {
+    echo "$0: invalid stream name."
+    exit 2
+}
+
+# Sorry, but this was the quickest and easiest way to do it and it works well
+# enough for our purposes...
+if [ $(lsb_release -cs) != "jessie" ]; then
+    exec logger -T -n syslog -P 514 -t "$stream"
```
chriskuehl: i should have commented though, this is a great find

once we're fully on stretch this script will become trivial, which is lovely

kpengboy: Which is why I thought of converting it to shell...

kpengboy:
On second thought, maybe we should backport `bsdutils` and install that in the containers instead? It would be much better efficiency and we might just be able to enable the syslogs for rt again...

chriskuehl:
I imagine we can just convert the services to use stretch? Virtually all of them are already I think (e.g. ocfweb), RT should be trivial since it's such a simple service.

kpengboy:
I suppose, although the soft freeze hasn't even happened yet...

chriskuehl:
That was the whole reason we wanted to get on stretch now (see the RT ticket). Last time we waited too long and couldn't get any of our fixes in before the freeze, which is why we have a bunch of patched packages in our apt repo for jessie even though every one of my patches was accepted.

If things in our containers break because of changes in testing, then either they'll break at the build stage, which is good, or we have insufficient tests and should fix that instead.

jvperrin:
Alright, now that all the services are on stretch, do we want to use `bsdutils` or?

chriskuehl:
@jvperrin sounds good to me, we can probably replace this with a one-liner that just execs `logger` with the appropriate params

the only complication i was thinking of is [rt#5699](https://rt.ocf.berkeley.edu/Ticket/Display.html?id=5699) -- but i think the actual fix to this is on the syslog server side (i.e. we should log the host that actually sends the logs). i confirmed that stretch loggers are sending the docker container id, which i think is ideal (since we can get the hostname of the host (jaws, etc.) from the ip.

kpengboy:
I'm superseding this with #55.

PR #29, opened 2016-12-22
Add fix-perms script

kpengboy:
The idea is to run this script from cron on the NFS server.

### DIFF
--- a/filehost/fix-perms
+++ b/filehost/fix-perms
@@ -3,11 +3,12 @@ set -euo pipefail
 
 echo "The following files were group or world writable:"
 
-# Find group/world-writable files in /home and /services/http owned by group
-# ocf. Remove the group- and world-writable bits on these files.
+# Find files in /home and /services/http which are world-writable or
+# group-writable and owned by group ocf. Remove the group- and world-writable
+# bits on these files.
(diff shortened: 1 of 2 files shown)


chriskuehl:
some thoughts:

* this should log what it changes
* this is very difficult to test, i think it needs a dry-run mode
* this should probably not go in utils since it's a cronjob (per the README). the only exceptions so far are mostly stuff used in docker.
* this should probably notify people whose files it is changing so that they know to not do that in the future (and can fix anything which is routinely creating files with bad permissions). obviously it should batch per user and not send 10k emails if it fixes 10k files :)

i don't really think this is the correct approach for this problem. i actually started an alternative about a week ago. it's fairly simple but also allows defining rules (e.g. it knows about `wp-config.php` files and will make sure they are not group/world-readable). i've got a flight tonight so i'll use the time to polish this up and share it.

daradib:
Also keep in mind security concerns/mitigations. For example, while `find -exec` should take care of properly escaping filenames (e.g., newlines), I think, there is a small [race condition](https://www.gnu.org/software/findutils/manual/html_node/find_html/Race-Conditions-with-_002dexec.html). `-execdir` might mitigate this.

PR #31, opened 2017-01-08
minutes: fix outstanding bugs

matthew-mcallister:
This should now be usable.
- Fix wrong file permissions and group ownership
- Fix string formatting
- Fix misc. logic


NickImpicciche:
I believe I'm going to re-implement string formatting with string templates, it avoids having to do exception handling. Will be trying to incorporate other changes in this pull request.

PR #32, opened 2017-01-10
Add volume control to ocf-tv

tahabi:
Now you can do `$ ocf-tv volume (0, 150)` to change the volume of the TV from 0% to 150% of baseline. It seems to work pretty decently, occasionally asks for you for a password to tornado. `connect` is still the default. 

When it does ask for a password however, `connect` will timeout if you aren't able to type the pwd fast enough, and after that fails, the terminal begins to exhibit weird behavior - no newlines on enter, no characters shown when typing, although it stills reads the input and executes when you press enter. Need to spawn a new terminal to fix it. Can't really explain why that happens when login fails.

PR #33, opened 2017-01-10
echo stty after ocf-tv connect

daradib:
If ssh is killed during a password prompt, terminal echo is
disabled and needs to be re-enabled. This should not happen
with cleanup after SIGTERM, but just in case, per #32.

Alternatively, we can disable ssh password prompts entirely
with `-o BatchMode=yes` and use Kerberos/GSSAPI.
Personally I prefer that approach to providing my password
to an ocf-tv script, but this is a simple change for now.

### DIFF
--- a/staff/lab/ocf-tv
+++ b/staff/lab/ocf-tv
@@ -40,11 +40,13 @@ def wait_for_port(host, port, timeout=5):
 
 def connect(args):
     port = unused_port()
-    proc = Popen(('ssh', '-L', '{}:localhost:5900'.format(port), '-N', args.host))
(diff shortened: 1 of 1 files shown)


chriskuehl:
yea i like requiring gssapi

tahabi:
This is pretty cool. Interesting to learn about echo not being re-enabled by ssh after cleaning up. I figure that is a bug in SSH?

daradib:
@tahabi ssh does clean up after itself (including reset terminal echo) when sent SIGTERM, but in this case it looks like ssh was still running (derp) because an exception was raised before calling terminate. Fixed in #34.

PR #39, opened 2017-02-01
volume control +/-, report and mute

tahabi:
now you can do 

    $ ocf-tv volume
    35%
    $ ocf-tv volume +30
    65%
    $ ocf-tv volume -20
    45%
    $ ocf-tv mute
    /

### DIFF
--- a/staff/lab/ocf-tv
+++ b/staff/lab/ocf-tv
@@ -48,18 +48,20 @@ def volume(args):
     if args.amount is not None:
         cmd += 'pactl set-sink-volume 0 {}%; '.format(args.amount)
 
-    # thanks SO for awk tricks
+    # awk prints the 9th line after the match
     cmd += "pactl list sinks | awk '/Sink #0/{nr[NR+9]}; NR in nr' | cut -d' ' -f 6;"
 
-    # tornado has 2 sinks, we default to sink 0 which is HDMI out to TV
     return call(('ssh', args.host, cmd))
 
 
 def volume_in_range(val):
+    # allow things like +20, -30, and absolute numbers
     x = int(val[1:]) if val[0] in ['+', '-'] else int(val)
     if x < 0 or x > 150:
         raise argparse.ArgumentTypeError('{} is not within the range [0, 150]'.format(val))
-    return '-- ' + val if val[0] is '-' else val  # pactl doesn't like negatives
+
+    # need -- before negatives or pactl interprets it as an arg
+    return '-- ' + val if val[0] == '-' else val
 
 
 def connect(args):
@@ -81,9 +83,11 @@ if __name__ == '__main__':
         'volume': volume,
         'mute': volume,
     }
+
     parser = argparse.ArgumentParser(
         description='Control the OCF television.',
     )
+
     parser.add_argument('-H', '--host', type=str, default='tv')
 
     command_subparser = parser.add_subparsers(dest='command')
@@ -93,10 +97,10 @@ if __name__ == '__main__':
     subparser_volume = command_subparser.add_parser('volume', help='set the volume on the TV using PulseAudio')
     subparser_volume.add_argument('amount', type=volume_in_range, nargs='?',
                                   help='volume in percent between 0% and 150%.'
-                                       '+/- a relative value is also allowed.')
+                                       ' +/- a relative value is also allowed.')
 
     subparser_mute = command_subparser.add_parser('mute', help='mute the TV')
-    # dirty hack so volume() can be reused
+    # hack so volume() can be reused
     subparser_mute.add_argument('--amount', default=0, help=argparse.SUPPRESS)
 
     if len(sys.argv) == 1:
--- a/staff/lab/ocf-tv
+++ b/staff/lab/ocf-tv
@@ -43,15 +43,23 @@ def wait_for_port(host, port, timeout=5, ssh_proc=None):
 
 
 def volume(args):
-    call(('ssh', args.host,
-          'export PULSE_SERVER=localhost; pactl set-sink-volume {} {}%'.format(args.sink, args.amount)))
+    cmd = 'export PULSE_SERVER=localhost; '
+
+    if args.amount is not None:
+        cmd += 'pactl set-sink-volume 0 {}%; '.format(args.amount)
+
+    # thanks SO for awk tricks
+    cmd += "pactl list sinks | awk '/Sink #0/{nr[NR+9]}; NR in nr' | cut -d' ' -f 6;"
+
+    # tornado has 2 sinks, we default to sink 0 which is HDMI out to TV
+    return call(('ssh', args.host, cmd))
 
 
 def volume_in_range(val):
-    x = int(val)
+    x = int(val[1:]) if val[0] in ['+', '-'] else int(val)
     if x < 0 or x > 150:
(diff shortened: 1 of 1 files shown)


staff/lab/ocf-tv
```diff
@@ -43,15 +43,23 @@ def wait_for_port(host, port, timeout=5, ssh_proc=None):
...
 
 def volume(args):
-    call(('ssh', args.host,
-          'export PULSE_SERVER=localhost; pactl set-sink-volume {} {}%'.format(args.sink, args.amount)))
+    cmd = 'export PULSE_SERVER=localhost; '
+
+    if args.amount is not None:
+        cmd += 'pactl set-sink-volume 0 {}%; '.format(args.amount)
+
+    # thanks SO for awk tricks
```
matthew-mcallister: Replace this comment with the one below, please.

staff/lab/ocf-tv
```diff
@@ -43,15 +43,23 @@ def wait_for_port(host, port, timeout=5, ssh_proc=None):
...
+    return call(('ssh', args.host, cmd))
 
 
 def volume_in_range(val):
-    x = int(val)
+    x = int(val[1:]) if val[0] in ['+', '-'] else int(val)
     if x < 0 or x > 150:
         raise argparse.ArgumentTypeError('{} is not within the range [0, 150]'.format(val))
-    return val
+    return '-- ' + val if val[0] is '-' else val  # pactl doesn't like negatives
```
matthew-mcallister: This is bizarre. What's wrong with `str(abs(val))`?

matthew-mcallister: Also, comments go on their own line.

staff/lab/ocf-tv
```diff
@@ -70,21 +78,26 @@ def connect(args):
...
-    subparser_volume.add_argument('amount', type=volume_in_range,
-                                  default='100', help='volume in percent between 0% and 150%')
+    subparser_volume = command_subparser.add_parser('volume', help='set the volume on the TV using PulseAudio')
+    subparser_volume.add_argument('amount', type=volume_in_range, nargs='?',
+                                  help='volume in percent between 0% and 150%.'
+                                       '+/- a relative value is also allowed.')
+
+    subparser_mute = command_subparser.add_parser('mute', help='mute the TV')
+    # dirty hack so volume() can be reused
+    subparser_mute.add_argument('--amount', default=0, help=argparse.SUPPRESS)
```
matthew-mcallister: Make the volume function take explicit arguments instead of the argparse return value.

staff/lab/ocf-tv
kpengboy: I think he means that if you pass `pactl` an argument that begins with a negative sign, `pactl` thinks it's an option, so you need to add the `--` to make `pactl` recognize it as an argument.

kpengboy: On another note, _never_ use `is` to compare strings. Always use `==`.

staff/lab/ocf-tv
```diff
@@ -43,15 +43,23 @@ def wait_for_port(host, port, timeout=5, ssh_proc=None):
...
 def volume(args):
-    call(('ssh', args.host,
-          'export PULSE_SERVER=localhost; pactl set-sink-volume {} {}%'.format(args.sink, args.amount)))
+    cmd = 'export PULSE_SERVER=localhost; '
+
+    if args.amount is not None:
+        cmd += 'pactl set-sink-volume 0 {}%; '.format(args.amount)
+
+    # thanks SO for awk tricks
+    cmd += "pactl list sinks | awk '/Sink #0/{nr[NR+9]}; NR in nr' | cut -d' ' -f 6;"
```
kpengboy: It'd be nice if we could find the proper way to print out this single field using `awk`.

staff/lab/ocf-tv
matthew-mcallister: I know what the code is supposed to do, it's just got bad readability (though you're right `str(abs(val))` is wrong, I misread the code, unsurprisingly).

staff/lab/ocf-tv
kpengboy: I suggest you start by understanding how the `awk` code you found works (and how `awk` works in general).

tahabi: The only part of this that I found on SO was using nr[NR+x] to get some line number after the match. Beyond that adding numerous variations of what should work in awk (`{print $6}`) doesn't seem to work, and all the examples I could find suggest to just farm it out to cut.

staff/lab/ocf-tv
tahabi: doing this breaks the general pattern used in most of the other scripts

staff/lab/ocf-tv
```diff
@@ -43,15 +43,25 @@ def wait_for_port(host, port, timeout=5, ssh_proc=None):
...
 def volume(args):
-    call(('ssh', args.host,
-          'export PULSE_SERVER=localhost; pactl set-sink-volume {} {}%'.format(args.sink, args.amount)))
+    cmd = 'export PULSE_SERVER=localhost; '
+
+    if args.amount is not None:
+        cmd += 'pactl set-sink-volume 0 {}%; '.format(args.amount)
+
+    # awk prints the 9th line after the match
+    cmd += "pactl list sinks | awk '/Sink #0/{nr[NR+9]}; NR in nr' | cut -d' ' -f 6;"
```
chriskuehl: i have no idea from reading the code why i would want the 9th line after the match. what are you actually computing here?

tahabi: Oh, just the way pactl prints the output, the volume is listed on the 9th line after the Sink name. I'll document that.

https://i.fluffy.cc/p1tmXBC7KfLCJwG2FvRpcrbD31HkKLrt.png

matthew-mcallister: Is it possible to do a slightly better match, e.g., find the first instance of "Volume:" after "Sink #0"? This makes the script resistant to changes in pactl output.

staff/lab/ocf-tv
```diff
@@ -43,15 +43,25 @@ def wait_for_port(host, port, timeout=5, ssh_proc=None):
...
+
+    return call(('ssh', args.host, cmd))
 
 
 def volume_in_range(val):
-    x = int(val)
+    # allow things like +20, -30, and absolute numbers
+    x = int(val[1:]) if val[0] in ['+', '-'] else int(val)
     if x < 0 or x > 150:
```
matthew-mcallister: No, really, you don't have to do string matching. Try this (acknowledging @kpengboy's remark):
```
x = int(val)
if abs(x) < 150:
 raise ValueError('Volume out of bounds.')
return '-- ' + str(x)
```
Also, this function name is totally misleading.

staff/lab/ocf-tv
matthew-mcallister: For the record, this is only worthwhile to do as long as it's easy, but I think awk should let you do that.

tahabi: Awk should, but I've been experimenting with different things from the awk manual and it doesn't seem to be working. I'll keep trying.

matthew-mcallister: You can hold off for now and just document what it does as-is. It may break anyways when we upgrade to stretch.

staff/lab/ocf-tv
tahabi: that makes a lot more sense.

PR #41, opened 2017-02-06
makevm: use predictable network interface names

matthew-mcallister:
Since we currently install stretch on new VMs, there should be no trouble merging.

### DIFF
--- a/staff/sys/makevm
+++ b/staff/sys/makevm
@@ -129,21 +129,32 @@ def get_ssh_connection(mac, user, password):
         time.sleep(1)
 
 
+def get_net_iface(ssh):
+    """Return the first available ethernet interface on the newly created VM.
+    In current versions of systemd/udev, they are supposed to be named starting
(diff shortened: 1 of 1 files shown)


jvperrin:
Nice, looks good to me, maybe it could support jessie too if it looked for anything starting with `eth` too, since that won't happen with systems interface names?

matthew-mcallister:
It could, I figured it just wouldn't get used much/ever---code minimalism, you know. Do you think it would be helpful?

chriskuehl:
nice, lgtm

if you *do* care about jessie, it's probably as simple as changing `iface.startswith('en')` to `iface.startswith(('en', 'eth'))`

matthew-mcallister:
Yep, I'm just being stubborn.

jvperrin:
Fair enough, well if we need to support any new jessie VMs, at least it's an easy change with this.

PR #48, opened 2017-05-01
migrate-vm: Support volume groups other than 'vg'

matthew-mcallister:
I tested this by moving dev-maelstrom to riptide and then back to jaws. Finally production VMs can be ported over there.

The script now reads the VM's LVM volume from virsh, and lets you specify the destination VG with the `--vg` option.

### DIFF
--- a/staff/sys/migrate-vm
+++ b/staff/sys/migrate-vm
@@ -38,14 +38,30 @@ import paramiko
 from ocflib.misc.shell import bold
 from ocflib.misc.shell import green
 from ocflib.misc.shell import red
+from ocflib.misc.shell import yellow
 
 
 HostAndVm = namedtuple('HostAndVm', ['host', 'vm'])
 
 client = None
 
+log_verbosity = 0
 
-def exec_command(cmd, ignore_status_code=False):
+
+def log_command(cmd, verbosity=0):
+    if log_verbosity >= verbosity:
+        if isinstance(cmd, str):
+            print(yellow(cmd))
+        else:
+            print(yellow(' '.join(cmd)))
+
+    return cmd
+
+
+def exec_command(cmd, ignore_status_code=False, verbosity=0):
+    if log_verbosity >= verbosity:
+        print('(remote) ' + yellow(cmd))
+
     stdin, stdout, stderr = client.exec_command(cmd)
 
     if not ignore_status_code:
@@ -60,7 +76,7 @@ def copy_disk_image(oldpath, newpath, num_bytes):
     _, stdout, _ = exec_command('sh -c "dd if=' + oldpath + ' bs=32M | pigz"', ignore_status_code=True)
 
     proc = Popen(
-        'gunzip | pv -s ' + str(num_bytes) + ' | dd of=' + newpath + ' bs=32M',
+        log_command('gunzip | pv -s ' + str(num_bytes) + ' | dd of=' + newpath + ' bs=32M'),
         stdin=PIPE,
         shell=True,
     )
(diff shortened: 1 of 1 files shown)


staff/sys/migrate-vm
```diff
@@ -107,11 +106,11 @@ def shut_down_vm(vm):
             break
 
         print('waiting for shutdown...')
-        time.sleep(1)
+        time.sleep(3)
```
matthew-mcallister: This is superficial, but how many times do you really need to see `waiting for shutdown` get printed?

staff/sys/migrate-vm
```diff
@@ -45,23 +42,25 @@ from ocflib.misc.shell import red
 
 HostAndVm = namedtuple('HostAndVm', ['host', 'vm'])
 
+client = None
+
 
 def exec_command(cmd, ignore_status_code=False):
-    print(bold(cmd))
```
matthew-mcallister: The script doesn't print out any of the significant commands it calls as it is, so I deleted this line. If anyone cares they can add a `-v` option.

staff/sys/migrate-vm
```diff
@@ -45,23 +42,25 @@ from ocflib.misc.shell import red
...
+        if exit_status != 0:
+            raise RuntimeError(bold(red('Command exited nonzero ({}): {}'.format(exit_status, cmd))))
 
     return stdin, stdout, stderr
 
 
-def copy_disk_image(vm, num_bytes):
-    _, stdout, _ = client.exec_command('sh -c "dd if=/dev/vg/' + vm + ' bs=32M | pigz"')
+def copy_disk_image(oldpath, newpath, num_bytes):
+    _, stdout, _ = exec_command('sh -c "dd if=' + oldpath + ' bs=32M | pigz"', ignore_status_code=True)
```
matthew-mcallister: If `ignore_status_code=False`, then `exec_command` blocks until the command returns, which is not what we want since that would result in a 15GB+ pipe buffer. The status code is ignored in the current script, but it could be checked after the whole transfer finishes.

staff/sys/migrate-vm
```diff
@@ -76,25 +75,25 @@ def copy_disk_image(vm, num_bytes):
...
 
     original = stdout.read().decode('ascii').strip()
     print(bold('Original SHA1 (on remote server): ') + original)
 
     copied = proc.stdout.read().decode('ascii').strip()
     print(bold('New SHA1 (on this server): ') + copied)
 
-    if original != copied:
+    # Pull out the hash string since file names may not match
+    if original.split()[1] != copied.split()[1]:
```
matthew-mcallister: I did worry about that and originally regex'd the two strings, but then I thought "Well, it's printed right above, won't someone notice if the output isn't a SHA1 sum?"

But I think you're right. If it says "success", there's the chance someone will miss some subtle problem with the output, or will just skip reading the hashes altogether.

staff/sys/migrate-vm
```diff
@@ -131,18 +136,23 @@ if __name__ == '__main__':
 
     parser = argparse.ArgumentParser(description=__doc__, formatter_class=argparse.RawDescriptionHelpFormatter)
     parser.add_argument('host_and_vm', metavar='host:vm', type=host_and_vm, help='Host and VM name to transfer.')
+    parser.add_argument('--vg', nargs='?', default='vg', help='Sets the destination volume group on the new host.')
```
matthew-mcallister: You're right; this was original not a `--` option but I made it one.

staff/sys/migrate-vm
matthew-mcallister: So, I do agree, and I was going to add `-v` and `-vv` (or probably `-quiet` and `-debug`) to print state-changing commands and all commands, respectively, as a PR. I inserted the stupid "I'm too cool to do it myself" tone to this comment I guess to save myself more explanation.

The command as it is prints absolutely none of the state-changing commands it performs on either host (except `virsh shutdown`), and no commands at all on the current host, which is pretty silly. It basically just needs a bunch of print statements inserted for the local commands in addition to this one here, but I thought I'd add log levels too since not all the commands called have any visible effect.

staff/sys/migrate-vm
```diff
@@ -151,29 +161,35 @@ if __name__ == '__main__':
...
+    num_bytes = get_lv_size(src_lvpath)
 
     print(bold('Creating logical volume here...'))
-    check_call(['lvcreate', '-L', str(num_bytes) + 'B', '-n', vm, '/dev/vg'])
+    check_call(['lvcreate', '-L', str(num_bytes) + 'B', '-n', vm, dest_vg])
 
     print(bold('Copying virsh definition...'))
     f = NamedTemporaryFile(mode='wb', delete=False)
-    f.write(get_definition(vm))
+    f.write(get_definition(vm).replace(bytes(src_lvpath, 'ascii'), bytes(dest_lvpath, 'ascii')))
```
matthew-mcallister: Thanks.

matthew-mcallister:
I added more consistent logging. It would be nice if we could add some primitive logging functionality to ocflib in the future.

matthew-mcallister:
Alright, I guess we can settle that the next time it comes up.

PR #49, opened 2017-05-01
firewall: use NetX

kpengboy:
This modifies the firewall script to use the free NetX JNLP
launcher rather than Oracle's proprietary JRE.

### DIFF
--- a/staff/sys/firewall
+++ b/staff/sys/firewall
@@ -1,9 +1,10 @@
 #!/bin/bash
-# Download and launch Java firewall config client
-cd $(mktemp -d)
-curl -sL http://contrib.ocf.berkeley.edu/jre-8u101-linux-x64.tar.gz | tar xfz -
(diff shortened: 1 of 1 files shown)


staff/sys/firewall
```diff
@@ -1,9 +1,10 @@
 #!/bin/bash
+set -euo pipefail
+
+TMPDIR="$(mktemp -d)"
+trap 'cd / && rm -rf "$TMPDIR"' EXIT
+cd "$TMPDIR"
```
matthew-mcallister: Why not just leave the temp dir after exiting?

kpengboy: I personally believe that apps should clean up their temporary files whenever possible. I believe that the autocleaning job should only be a backup option for those files which could not otherwise be cleaned.

matthew-mcallister: Suit yourself.

jvperrin: (Said the lazy tailor)

chriskuehl: lmao

PR #50, opened 2017-06-14
makevm: use SSH key instead of password

kpengboy:


staff/sys/makevm
```diff
@@ -153,10 +153,26 @@ def configure_network(ssh, ipv4):
         '\tgateway {}'.format(OCF_GATEWAY_V6),
     ]
 
-    ssh.exec_command('echo -n "" > /etc/network/interfaces')
-
+    ssh.exec_command('> /etc/network/interfaces')
```
matthew-mcallister: Nice

staff/sys/makevm
```diff
@@ -153,10 +153,26 @@ def configure_network(ssh, ipv4):
...
+        ssh.exec_command("echo '{}' >> /etc/network/interfaces".format(line))
+
+
+def wait_drop_into_shell(ip, user, key_file):
+    client = paramiko.SSHClient()
+    client.set_missing_host_key_policy(paramiko.AutoAddPolicy())
+
+    # wait for the VM to be live
+    print('Waiting for VM to finish rebooting...')
+    while True:
```
jvperrin: I feel like making this _not_ an infinite loop would be better, maybe it could exit once paramiko is able to connect so that the next ssh command could be run outside of the loop? Also it would be good to print what SSH command is being run so that if someone disconnects from the single SSH session they are dropped into they know how to get back, since it will require a key now.

staff/sys/makevm
chriskuehl: well, it's certainly less clear what it's doing now...

chriskuehl: (in other words, my vote is that this change harms readability for no benefit)

PR #52, opened 2017-08-20
Add shellcheck pre-commit hook

jvperrin:
A bit messy on the first commit because there was a lot to fix, but `shellcheck` should keep our shell scripts more consistent and less error-prone. I'd like to submit a PR upstream to [pre-commit-shell](https://github.com/detailyang/pre-commit-shell) to support using types, because currently it only supports checking files that end in the correct extension:

    files: (\.sh|\.zsh|\.ksh)$

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -41,5 +41,4 @@
         description: Check Shell Syntax on ALL staged files with user friendly messages and colors
         entry: shell-lint.sh
         language: script
-        exclude: ^.deactivate.sh$
         types: [executable, shell]
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -31,3 +31,15 @@
     hooks:
     -   id: remove-tabs
         language_version: python2.7
+# TODO: Replace this with the actual repo
+# (https://github.com/detailyang/pre-commit-shell)
+# once it supports types and not just files ending in shell extensions
+-   repo: local
+    hooks:
+    -   id: shell-lint
+        name: Shell Syntax Check
+        description: Check Shell Syntax on ALL staged files with user friendly messages and colors
+        entry: shell-lint.sh
+        language: script
+        exclude: ^.deactivate.sh$
+        types: [executable, shell]
--- a/acct/check
+++ b/acct/check
@@ -13,7 +13,7 @@ wtmp_files='/var/log/wtmp /var/log/wtmp.1'
 login_count=5
 
 usage(){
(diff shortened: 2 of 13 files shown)


matthew-mcallister:
Wow, this is like 10+ years of code getting linted here.

acct/check
```diff
@@ -145,7 +145,7 @@ echo
 paper view "$user"
 
 # check User_Info entries
-User_Info=$([ -f "$User_Info_file" ] && grep ^$user: "$User_Info_file" | cut -d: -f2-)
+User_Info=$([ -f "$User_Info_file" ] && grep ^"$user": "$User_Info_file" | cut -d: -f2-)
```
chriskuehl: this works but is kinda weird. i'd write `"^$user"` instead of `^"$user"`

.pre-commit-config.yaml
```diff
@@ -31,3 +31,15 @@
...
+# (https://github.com/detailyang/pre-commit-shell)
+# once it supports types and not just files ending in shell extensions
+-   repo: local
+    hooks:
+    -   id: shell-lint
+        name: Shell Syntax Check
+        description: Check Shell Syntax on ALL staged files with user friendly messages and colors
+        entry: shell-lint.sh
+        language: script
+        exclude: ^.deactivate.sh$
```
chriskuehl: shouldn't need this anymore if using types

chriskuehl: oh, i skimmed too fast. yeah, you probably need this

chriskuehl: (i assumed you were excluding the .activate.sh symlink rather than the .deactivate.sh file)

jvperrin:
We do [install `shellcheck` in extrapackages](https://github.com/ocf/puppet/blob/aef2a578f316c9eb09b624746b29eeb7f33543ed/modules/ocf/manifests/extrapackages.pp#L161), but yeah, it won't work everywhere, so that's a good point.

.pre-commit-config.yaml
jvperrin: Ah, good point, thanks. I put this there before using types, so that's better

acct/check
jvperrin: Haha yeah, that would be better, too many changes to make that I didn't look at this very closely

.pre-commit-config.yaml
jvperrin: Still seems to work without it, so that's fine, I'll remove it.

PR #53, opened 2017-08-20
Add dnspython to dev requirements

jvperrin:
This PR is also to test if PRs actually get run by Jenkins now or not.

### DIFF
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,2 +1,3 @@
+dnspython
 ocflib
 pre-commit


jvperrin:
Cool, PRs do get tested now, that's nice

jvperrin:
Yeah, it's a bit strange like that, I mostly wanted this here so that the scripts can be tested if the virtualenv is activated, but ideally we'd have something in here to make it so that packages don't have to be installed and the virtualenv can be used instead.

PR #55, opened 2017-09-03
stdin2syslog: directly exec logger

kpengboy:
Now that we're on stretch and this works

### DIFF
--- a/staff/sys/stdin2syslog
+++ b/staff/sys/stdin2syslog
@@ -2,13 +2,12 @@
 set -euo pipefail
 
 if [ $# -lt 1 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
-    echo "Usage: $0 stream
-
-Write stdin logs to OCF's remote syslog server.
-
-You can use arbitrary stream names (without configuring them before); each
-stream will show up at syslog:/var/log/remote/<stream>.log.
-"
+    echo "Usage: $0 stream"
+    echo
+    echo "Write stdin logs to OCF's remote syslog server."
+    echo
+    echo "You can use arbitrary stream names (without configuring them before); each"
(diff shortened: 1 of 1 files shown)


staff/sys/stdin2syslog
```diff
@@ -1,49 +1,22 @@
...
-def main(argv=None):
-    parser = argparse.ArgumentParser(
-        description=__doc__,
-        formatter_class=argparse.RawDescriptionHelpFormatter,
-    )
-    parser.add_argument('stream')
-    args = parser.parse_args(argv)
+You can use arbitrary stream names (without configuring them before); each
+stream will show up at syslog:/var/log/remote/<stream>.log.
+"
```
jvperrin: Not a huge fan of the indentation here, but not sure what would be better, maybe a here doc or something?

matthew-mcallister: Usually you would either save it to a variable at top level indentation or just use multiple `echo` commands.

kpengboy:
I decided to switch the script to bash after trying versions in both languages. I ultimately thought that, for this particular script, the bash version was overall clearer, as it had less syntax getting in the way of communicating the main idea of what the script does.

staff/sys/stdin2syslog
```diff
@@ -1,49 +1,22 @@
...
-import subprocess
-import sys
-
-
-VALID_STREAM_NAME = re.compile('^[a-z0-9\-_]+$')
-
-
-class LoggingFailure(Exception):
-    pass
+if [ $# -lt 1 ] || [ "$1" = "-h" ] || [ "$1" = "--help" ]; then
```
matthew-mcallister: It just occurred to me: you might want to check `$# -ne 1` instead since the script expects exactly one argument.

kpengboy: Good point.

PR #56, opened 2017-09-04
check: show dns for primary vhost

abizer:
shellcheck cannot be appeased at my current level of experience in writing portable shell scripts

please provide feedback on anything that could be done better in this, if possible.

+= checking for mail vhosts

### DIFF
--- a/acct/check
+++ b/acct/check
@@ -153,35 +153,13 @@ if [ -f "$vhost_mail_file" ] && vhost_mail=$(grep -E "^[#[:space:]]*$user[![:spa
 fi
 
 # check dns entries
-expand_fqdns()
-{
-  while read -r entry; do
-    if [ -n "$entry" ]; then
-      username=$(echo "$entry" | cut -d' ' -f1)
-      domain=$(echo "$entry" | cut -d' ' -f2)
+dns=$(./check-dns "$user")
 
-      if [ "$domain" = "-" ]; then
-        echo "$username.berkeley.edu"
-      elif echo "$domain" | grep -qE '.+\.[a-z]+$'; then
-        echo "$domain"
-      else
-        echo "$domain.berkeley.edu"
-      fi
-    fi
-  done
-}
-
-fqdns=$(printf "%s\n%s\n%s\n" "$vhost" "$vhost_app" "$vhost_mail" | grep -v '#' | expand_fqdns)
-
-if [ -n "$fqdns" ]; then
+if [ -n "$dns" ]; then
   echo
   echo "DNS records for $user:"
   echo "-------------------------------------------------------------------------------"
-
-  echo "$fqdns" | sort -d | uniq | while read -r fqdn; do
-    dig A "$fqdn" AAAA "$fqdn" MX "$fqdn" CNAME "$fqdn" 2>&1 | grep -E "^$fqdn" | sort -d | uniq
-  done
-
+  echo "$dns"
   echo "-------------------------------------------------------------------------------"
 fi
 
--- a/acct/check
+++ b/acct/check
@@ -153,7 +153,7 @@ if [ -f "$vhost_mail_file" ] && vhost_mail=$(grep -E "^[#[:space:]]*$user[![:spa
 fi
 
 # check dns entries
-dns=$(./check-dns "$user")
+dns=$("$(dirname "$(readlink -f -- "$0")")"/check-dns "$user")
 
 if [ -n "$dns" ]; then
   echo
--- a/acct/check
+++ b/acct/check
@@ -1,5 +1,4 @@
 #!/bin/sh
-# shellcheck shell=sh
 
 # Look up user for relevant account information.
 
@@ -177,28 +176,23 @@ expand_fqdns()
   done
 }
 
-# shellcheck disable=SC2028
-fqdns=$(echo "$vhost\n$vhost_app\n$vhost_mail" | grep -v '#' | expand_fqdns)
+fqdns=$(printf "%s\n%s\n%s" "$vhost" "$vhost_app" "$vhost_mail" | grep -v '#' | expand_fqdns)
 
 if [ -n "$fqdns" ]; then
   echo
   echo "DNS records for $user:"
   echo "-------------------------------------------------------------------------------"
 
-  # shellcheck disable=SC2028
   echo "$fqdns" | sort -d | uniq | while read -r fqdn; do
-    dns=$(dig A "$fqdn" AAAA "$fqdn" MX "$fqdn" CNAME "$fqdn" 2>&1 | grep -E "^$fqdn" | sort -d | uniq)
-    echo "$dns"
+    dig A "$fqdn" AAAA "$fqdn" MX "$fqdn" CNAME "$fqdn" 2>&1 | grep -E "^$fqdn" | sort -d | uniq
   done
 
   echo "-------------------------------------------------------------------------------"
 fi
 
-
 # check print quota
 echo
 paper view "$user"
-
 
 # check User_Info entries
 User_Info=$([ -f "$User_Info_file" ] && grep "^$user": "$User_Info_file" | cut -d: -f2-)
@@ -215,7 +209,6 @@ echo
 echo "Signatories/Signatory for:"
 signat user "$user"
 
-
 # check user crontab, probably requires root privileges
 crontab=$(crontab -lu "$user" 2> /dev/null | sed '/^[ ]*$/d')
 if [ -n "$crontab" ]; then
@@ -225,7 +218,6 @@ if [ -n "$crontab" ]; then
   echo "$crontab"
   echo "-------------------------------------------------------------------------------"
 fi
-
 
 # check currently running processes in tree format
 ps=$(ps Sfw -u "$user" 2> /dev/null | tail -n +2)
--- a/acct/check
+++ b/acct/check
@@ -1,4 +1,6 @@
 #!/bin/sh
+# shellcheck shell=sh
+
 # Look up user for relevant account information.
 
 # specify location of User_Info file
@@ -7,6 +9,8 @@ User_Info_file='/home/s/st/staff/User_Info'
 vhost_file='/home/s/st/staff/vhost/vhost.conf'
 # specify location of vhost-app file
 vhost_app_file='/home/s/st/staff/vhost/vhost-app.conf'
+# specify location of vhost-mail file
+vhost_mail_file='/home/s/st/staff/vhost/vhost-mail.conf'
 # specify location of wtmp files
 wtmp_files='/var/log/wtmp /var/log/wtmp.1'
 # default number of logins to display from wtmp files
@@ -122,6 +126,7 @@ if [ -n "$calnetuid" ]; then
   fi
 fi
 
+
 # check virtual hosting config
 if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$vhost_file"); then
   echo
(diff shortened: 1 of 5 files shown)


acct/check
```diff
@@ -132,18 +137,69 @@ if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$v
...
+    elif [ "$(echo $domain | grep -E '*\.[a-z]+$')" ]; then
+      echo "$domain"
+    else
+      echo "$domain.berkeley.edu"
+    fi
+  done
+}
+
+# shellcheck disable=SC2028
+fqdns=$(echo "$vhost\n$vhost_app\n$vhost_mail" | grep -v '#' | expand_fqdns)
```
tahabi: this seems ugly to me (and shellcheck doesn't like it) but all the alternatives I tried don't seem to work or fail on certain cases.

kpengboy: Have you tried using `printf "%s\n%s\n%s\n" "$vhost" "$vhost_app" "$vhost_mail"`?

btw the way you have it now wouldn't work on bash...

acct/check
```diff
@@ -132,18 +137,69 @@ if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$v
...
+
+if [ -n "$fqdns" ]; then
+  echo
+  echo "DNS records for $user:"
+  echo "-------------------------------------------------------------------------------"
+
+  # shellcheck disable=SC2028
+  echo "$fqdns" | sort -d | uniq | while read -r fqdn; do
+    dns=$(dig A "$fqdn" AAAA "$fqdn" MX "$fqdn" CNAME "$fqdn" 2>&1 | grep -E "^$fqdn" | sort -d | uniq)
```
kpengboy: Do you have to save this output to a variable?

acct/check
```diff
@@ -1,4 +1,6 @@
 #!/bin/sh
+# shellcheck shell=sh
```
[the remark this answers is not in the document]
tahabi: yes, I was hoping it would kill some of the errors

acct/check
```diff
@@ -122,6 +126,7 @@ if [ -n "$calnetuid" ]; then
   fi
 fi
 
+
```
matthew-mcallister: Why a double newline here? More further down too.

tahabi: fixed

tahabi:
any further comments on this?

acct/check
```diff
@@ -132,11 +136,57 @@ if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$v
...
+      echo "$username.berkeley.edu"
+    elif [ "$(echo $domain | grep -E '*\.[a-z]+$')" ]; then
+      echo "$domain"
+    else
+      echo "$domain.berkeley.edu"
+    fi
+  done
+}
+
+fqdns=$(printf "%s\n%s\n%s" "$vhost" "$vhost_app" "$vhost_mail" | grep -v '#' | expand_fqdns)
```
kpengboy: You need to add a final `\n` to the `printf` format string, or else `expand_fqdns` is going to miss the last line in the `vhost_mail` entries

acct/check
```diff
@@ -132,11 +136,57 @@ if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$v
...
+      continue
+    fi
+    
+    username=$(echo "$entry" | cut -d' ' -f1)
+    domain=$(echo "$entry" | cut -d' ' -f2)
+
+    # shellcheck disable=SC2143,SC2086,SC2063
+    if [ "$domain" = "-" ]; then
+      echo "$username.berkeley.edu"
+    elif [ "$(echo $domain | grep -E '*\.[a-z]+$')" ]; then
```
jvperrin: To fix [SC2143](https://github.com/koalaman/shellcheck/wiki/SC2143), [SC2086](https://github.com/koalaman/shellcheck/wiki/SC2086), and [SC2063](https://github.com/koalaman/shellcheck/wiki/SC2063), this should just be:

```
elif echo "$domain" | grep -qE '.+\.[a-z]+$'; then
```

Also, if the goal here is to just check if the domain contains a dot, then you can make it even simpler and just do something like `grep -q '\.'` instead.

acct/check
```diff
@@ -132,11 +136,57 @@ if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$v
...
+
+fqdns=$(printf "%s\n%s\n%s" "$vhost" "$vhost_app" "$vhost_mail" | grep -v '#' | expand_fqdns)
+
+if [ -n "$fqdns" ]; then
+  echo
+  echo "DNS records for $user:"
+  echo "-------------------------------------------------------------------------------"
+
+  echo "$fqdns" | sort -d | uniq | while read -r fqdn; do
+    dig A "$fqdn" AAAA "$fqdn" MX "$fqdn" CNAME "$fqdn" 2>&1 | grep -E "^$fqdn" | sort -d | uniq
```
jvperrin: Nice, this will be quite useful to have

acct/check
tahabi: nice

acct/check
tahabi: I had disabled this check because using `grep -qE` was causing this clause to trigger incorrectly in certain cases but it turns out it was just bad bash on my part. This is fixed now and no more shellcheck problems.

acct/check
```diff
@@ -132,11 +135,53 @@ if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$v
...
+
+      if [ "$domain" = "-" ]; then
+        echo "$username.berkeley.edu"
+      elif echo "$domain" | grep -qE '.+\.[a-z]+$'; then
+        echo "$domain"
+      else
+        echo "$domain.berkeley.edu"
+      fi
+    fi
+  done
```
chriskuehl: :(

could we put this new stuff in a python script and call it? it kinda sucks to be re-implementing the parsing logic

jvperrin: That's true, it would be nice to have the same parsing code for this too...

tahabi: I switched this out for a python script

abizer:
any further comments on this?

acct/check-dns
```diff
@@ -0,0 +1,52 @@
...
+            for rdata in resolver.query(domain, 'CNAME', raise_on_no_answer=False):
+                print("{}\tIN\tCNAME\t{}".format(domain, rdata))
+
+            for rdata in resolver.query(domain, 'MX', raise_on_no_answer=False):
+                print("{}\tIN\tMX\t{} {}".format(domain, rdata.preference, rdata.exchange))
+        except Exception as e:
+            print("[{}]: {}".format(domain, e))
+            
+if __name__ == '__main__':
+    if len(sys.argv) < 2:
```
chriskuehl: what if it's more? probably want to assert it's exactly 2

acct/check-dns
```diff
@@ -0,0 +1,52 @@
...
+            for rdata in resolver.query(domain, 'AAAA', raise_on_no_answer=False):
+                print("{}\tIN\tAAAA\t{}".format(domain, rdata))
+
+            for rdata in resolver.query(domain, 'CNAME', raise_on_no_answer=False):
+                print("{}\tIN\tCNAME\t{}".format(domain, rdata))
+
+            for rdata in resolver.query(domain, 'MX', raise_on_no_answer=False):
+                print("{}\tIN\tMX\t{} {}".format(domain, rdata.preference, rdata.exchange))
+        except Exception as e:
+            print("[{}]: {}".format(domain, e))
```
chriskuehl: just let the exception raise? this is almost always an antipattern that destroys information you actually want when debugging

if there are expected errors here, they should probably be handled individually

acct/check-dns
```diff
@@ -0,0 +1,52 @@
...
+
+
+def main(username):
+
+    domains = set()
+
+    for k, v in web.get_vhosts().items():
+        if v['username'] == username:
+            domains.add(k)
+            domains.update(v['aliases'])
```
chriskuehl: what are k/v? can you call these `domain_name` and `vhost_config` or something? i don't actually know what they are

acct/check-dns
```diff
@@ -0,0 +1,52 @@
...
+
+from dns import resolver
+
+from ocflib.vhost import web
+from ocflib.vhost import mail
+from ocflib.vhost import application
+
+
+def main(username):
+
```
chriskuehl: weird spacing

acct/check-dns
```diff
@@ -0,0 +1,52 @@
+#!/usr/bin/env python3
+"""Print DNS entries for domains associated with our users."""
+
+import sys
+
+from dns import resolver
+
+from ocflib.vhost import web
```
chriskuehl: why empty line between two 3rd party imports?

acct/check-dns
```diff
@@ -0,0 +1,52 @@
...
+                print("{}\tIN\tCNAME\t{}".format(domain, rdata))
+
+            for rdata in resolver.query(domain, 'MX', raise_on_no_answer=False):
+                print("{}\tIN\tMX\t{} {}".format(domain, rdata.preference, rdata.exchange))
+        except Exception as e:
+            print("[{}]: {}".format(domain, e))
+            
+if __name__ == '__main__':
+    if len(sys.argv) < 2:
+        print("Usage: check-dns <username>")
```
chriskuehl: lol, exit here instead of falling through and calling main?

abizer: oops

acct/check
```diff
@@ -132,11 +135,31 @@ if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$v
...
+if [ -f "$vhost_mail_file" ] && vhost_mail=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$vhost_mail_file"); then
+  echo
+  echo "Virtual mail:"
+  echo "-------------------------------------------------------------------------------"
+  echo "$vhost_mail"
+  echo "-------------------------------------------------------------------------------"
+fi
+
+# check dns entries
+dns=$(./check-dns "$user")
```
kpengboy: I think you'll want something like `"$(dirname "$(readlink -f -- "$0")")"/check-dns` here instead

abizer: what does this do differently besides call the script directly instead of through the symlink? I'm not sure I understand the difference?

kpengboy: Well, the way you have it now, it won't work after you deploy it unless your current directory happens to be `/opt/share/utils/bin`

abizer: oh, that makes sense. thanks

bin/shell-lint.sh
```diff
@@ -0,0 +1 @@
+../shell-lint.sh
```
kpengboy: Hmm...it looks like we'll have to change `update-symlinks` to exclude this. We might want to pass a `-mindepth` option to one of the `find` commands in that script.

kpengboy: For that matter `update-symlinks` ought to exclude itself too

acct/check
```diff
@@ -132,11 +135,31 @@ if [ -f "$vhost_file" ] && vhost=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$v
...
+if [ -f "$vhost_mail_file" ] && vhost_mail=$(grep -E "^[#[:space:]]*$user[![:space:]]" "$vhost_mail_file"); then
+  echo
+  echo "Virtual mail:"
+  echo "-------------------------------------------------------------------------------"
+  echo "$vhost_mail"
+  echo "-------------------------------------------------------------------------------"
+fi
+
+# check dns entries
+dns=$("$(dirname "$(readlink -f -- "$0")")"/check-dns "$user")
```
abizer: @kpengboy looks good?

kpengboy: Looks good, would you be able to fix update-symlinks too?

abizer: done

PR #57, opened 2017-09-05
stdin2syslog: use TCP again

kpengboy:
We used to use TCP in `stdin2syslog`, but we then switched to using UDP in an unsuccessful attempt to solve rt#5913. Shall we switch back to using TCP?


chriskuehl:
I'm still sort of uneasy about using TCP, as if the syslog server is broken it could potentially lock things up in very confusing ways. 

What's the motivation of this change? If we're just worried about dropping log lines, maybe we could add monitoring to the syslog server?

I think there's a reason most people run their syslog servers using UDP.

kpengboy:
Let's stay on UDP then, at least for now.

PR #58, opened 2017-09-06
Update meetings script in light of recent BoD eligibility changes

kpengboy:
We recently removed the regulation requiring people to attend BoD 4 times before joining, in light of the split between BoD and staff meetings. This patch updates the minutes taker to behave accordingly.


tahabi:
and this one too

Issue #44, opened 2017-02-21
minutes: warn if accounts in attendance list are nonexistent/not in staff


kpengboy:
This should prevent spelling errors from creeping in to the attendance list.

NickImpicciche:
Planning to do after my midterm

abizer:
@fydai there was a staff member who you mentioned was interested in something like this, perhaps tag them in this issue?

PR #61, opened 2017-09-07
Distinguish directors and guests in BoD attendance lists

kpengboy:
This commit makes the minutes script split the "Attendance"
block in BoD minutes. The note-taker still writes a single
Attendance block, like before; however, after they exit the
text editor, the Attendance block is replaced with two blocks,
an attendance block for directors and an attendance block for
guests. Guests are further marked with an asterisk if they
choose to join BoD at that meeting.
    
Major changes have been made to the BoD module. Most notably,
the membership file is eliminated. Membership is now dynamically
determined from the attendance lists in the minutes.
    
One drawback is that you cannot run `minutes` to take BoD minutes
twice in the same day without re-combining the attendance lists,
or else it crashes. This can be remedied in the future.

### DIFF
--- a/staff/meetings/bod
+++ b/staff/meetings/bod
@@ -3,34 +3,33 @@ import argparse
 import sys
 from os.path import join
 
+import bod
 import meetings
 
 
+def get_last_meeting():
+    """Returns a (semester, filename) pair for the most recent meeting."""
+    if not hasattr(get_last_meeting, 'cached_val'):
+        meeting_lst = meetings.get_minutes(bod.get_bod_minutes_path())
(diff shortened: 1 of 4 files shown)


kpengboy:
Comments? If not I'm going to merge these.

tahabi:
Perhaps we should check to make sure everyone who's "on bod" has an OCF account to capture occasions when the minutes-taker misspells usernames, and prompts them to correct it or mark them as a guest. Transcription errors are likely and may lead to people getting kicked off bod because of faulty minutes-taking.

kpengboy:
Yeah, that's issue #44

PR #62, opened 2017-09-13
Add setup wordpress script

kkuehler:
A script that an OCF user can run (ideally should be called in response to a registered user clicking a button on the OCF Website) to easily set up WordPress on their account and forgo all command line interaction. The user can finish setting up WordPress in a browser at https://www.ocf.berkeley.edu/~username

### DIFF
--- a/makeservices/easywp
+++ b/makeservices/easywp
@@ -1,0 +1,38 @@
+#!/bin/bash
+
+php_defines() {
+    sed -i "s/$1/$2/" "$WPCONFIG"
+}
+
+WEBROOT=$HOME/public_html
+WPCONFIG=wp-config.php
+USER=$(whoami)
+
+# Create symlink to webroot
+makehttp
+cd "$WEBROOT"
+rm -rf *
+
+# Get SQL password
+SQLPASS=$(echo "yes" | makemysql | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
+[ -z "$SQLPASS" ] && exit 1
+
+# Install WordPress
+wp core download
+cp wp-config-sample.php "$WPCONFIG"
+
+# dos2unix
+sed -i 's/\r//' "$WPCONFIG"
+
+# Modify the config file with correct info
+php_defines 'database_name_here' "$USER"
+php_defines 'username_here' "$USER"
+php_defines 'password_here' "$SQLPASS"
+php_defines 'localhost' 'mysql'
+php_defines 'utf8' 'utf8mb4'
+
+# Grab auth keys and salts from WP API
+sed -i '/put your unique phrase here/d' "$WPCONFIG"
+echo >> "$WPCONFIG"
+curl -sS -X GET https://api.wordpress.org/secret-key/1.1/salt/ >> "$WPCONFIG"
+
--- a/makeservices/easywp
+++ b/makeservices/easywp
@@ -1,5 +1,19 @@
 #!/bin/bash
 
+# Creates symlink to user's webroot. If the symlink exists and files
+# are present in webroot, they are backed up in the user's home directory.
+# A user's database is created or, if it already exists, the password 
+# is reset. WordPress is installed and the wp-config file is filled
+# out with the requisite information. Further configuration can be 
+# done by the user from a browser.  
+#
+# WARNING: This script should generally be run on a account without
+#          an existing website or database!
+#
+
+set -euxo pipefail
+shopt -s dotglob
+
 php_defines() {
     sed -i "s/$1/$2/" "$WPCONFIG"
 }
@@ -11,11 +25,24 @@ USER=$(whoami)
 # Create symlink to webroot
 makehttp
 cd "$WEBROOT"
-rm -rf *
+
+# If user has files, move them to a new location. 
+if [ -n "$(ls -A)" ]; then
+    TS=$(date +"%Y-%m-%d_%H:%M:%S")
+    BACKUP="$HOME/webroot-$TS"
+    mkdir -p "$BACKUP"
+    echo "Webroot not empty. Moving current files to $BACKUP..." 
+    mv ./* "$BACKUP"
+fi
 
 # Get SQL password
+echo "Resetting database..."
 SQLPASS=$(echo "yes" | makemysql | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
-[ -z "$SQLPASS" ] && exit 1
+if [ -z "$SQLPASS" ]; then
+    echo "  Error: Could not retrieve database password. Run makemysql to see the issue."
+    exit 1
+fi
+echo "SQL set up and the password can be found in $WPCONFIG ..."
 
 # Install WordPress
 wp core download
@@ -36,3 +63,4 @@ sed -i '/put your unique phrase here/d' "$WPCONFIG"
 echo >> "$WPCONFIG"
 curl -sS -X GET https://api.wordpress.org/secret-key/1.1/salt/ >> "$WPCONFIG"
 
+echo "WordPress installed successful. Go to https://www.ocf.berkeley.edu/~$USER to finish the setup process."
--- a/makeservices/easywp
+++ b/makeservices/easywp
@@ -14,7 +14,7 @@ shopt -s dotglob
 echo -e "\033[33mWARNING: This script should generally be run on a fresh account!\033[00m"
 echo -e "A user with an existing website or database will have their files
 backed up, but their database password will change and the WordPress
-install may clutter or try to migrate tables.\n"
+install may migrate and reuse existing tables.\n"
 
 read -rp "Are you sure you want to continue? [y/N] " idunno
 idunno=${idunno,,}
@@ -33,15 +33,16 @@ cd "$webroot"
 # Check if user has files and let user decide what they want to do.
 if [[ -n "$(ls -A)" ]]; then
     echo
-    echo -e "\033[33mThere are currently files in webroot!\033[00m\n"
+    echo -e "\033[33mThere are currently files in webroot! These will be backed up.\033[00m\n"
     read -rp "Do you want to continue? [y/N] " whoknows
 
     whoknows=${whoknows,,}
     if [[ "$whoknows" =~ ^(yes|y) ]]; then
         ts=$(date +"%Y-%m-%d_%H:%M:%S")
-        backup="$HOME/webroot-$ts"
+        backup="$HOME/public_html-$ts"
         mkdir -p "$backup"
         echo "Moving current files to $backup ..."
+        # Note that dotglob is set so hidden files are moved too
         mv ./* "$backup"
     else
         echo "Ok, bye!"
@@ -61,6 +62,6 @@ echo "SQL set up and the password can be found in wp-config.php ..."
 # Install WordPress and create config file
 wp core download
 wp config create --dbname="$user" --dbuser="$user" --dbpass="$sqlpass" --dbhost=mysql --dbcharset=utf8mb4
-# Somehow 644 is the default...
+# Somehow 644 is the default... See https://core.trac.wordpress.org/ticket/37264
 chmod 600 "$webroot/wp-config.php"
 echo "WordPress install successful. Go to https://www.ocf.berkeley.edu/~$user to finish the setup process."
--- a/makeservices/easywp
+++ b/makeservices/easywp
@@ -2,23 +2,23 @@
 
 # Creates symlink to user's webroot. If the symlink exists and files
 # are present in webroot, they are backed up in the user's home directory.
-# A user's database is created or, if it already exists, the password 
+# A user's database is created or, if it already exists, the password
 # is reset. WordPress is installed and the wp-config file is filled
-# out with the requisite information. Further configuration can be 
-# done by the user from a browser.  
+# out with the requisite information. Further configuration can be
+# done by the user from a browser.
 #
 
 set -euo pipefail
 shopt -s dotglob
 
-echo -e "\033[31mWARNING: This script should generally be run on a fresh account!\033[00m"
-echo -e "A user with an existing website or database will have their files 
-backed up, but their database password will change and the WordPress 
+echo -e "\033[33mWARNING: This script should generally be run on a fresh account!\033[00m"
+echo -e "A user with an existing website or database will have their files
+backed up, but their database password will change and the WordPress
 install may clutter or try to migrate tables.\n"
 
-read -rp "Are you sure you want to continue? [Y/n] " idunno
+read -rp "Are you sure you want to continue? [y/N] " idunno
 idunno=${idunno,,}
-if ! [[ "$idunno" =~ ^(yes|y|^$) ]]; then
+if ! [[ "$idunno" =~ ^(yes|y) ]]; then
     echo "Ok, bye!"
     exit 0
 fi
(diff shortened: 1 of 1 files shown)


makeservices/easywp
```diff
@@ -0,0 +1,38 @@
+#!/bin/bash
```
matthew-mcallister: Your commit message should be in a comment below this line.

matthew-mcallister: For good measure, you should also warn that this script only works on a fresh account, i.e. not one that has an existing website or database. Or you can bail out or show a prompt.

makeservices/easywp
```diff
@@ -0,0 +1,38 @@
...
+
```
matthew-mcallister: You also should start your script off with `set -euxo pipefail`.

makeservices/easywp
```diff
@@ -0,0 +1,38 @@
...
+USER=$(whoami)
+
+# Create symlink to webroot
+makehttp
+cd "$WEBROOT"
+rm -rf *
+
+# Get SQL password
+SQLPASS=$(echo "yes" | makemysql | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
+[ -z "$SQLPASS" ] && exit 1
```
matthew-mcallister: You should definitely have an error message here since you're soaking up stdout above.

makeservices/easywp
```diff
@@ -0,0 +1,38 @@
...
+}
+
+WEBROOT=$HOME/public_html
+WPCONFIG=wp-config.php
+USER=$(whoami)
+
+# Create symlink to webroot
+makehttp
+cd "$WEBROOT"
+rm -rf *
```
matthew-mcallister: You might want a more complicated operation here, as this will leave e.g. `.htaccess` lying around.

chriskuehl: Yeah... to be honest I'm not sure we ever want to run `rm -rf *` in a script like this.

I totally get why it's here, but I think it's likely to accidentally cause a lot of damage. I could even imagine staff running the script for testing without realizing it would delete their entire public_html/*.

matthew-mcallister: I agree, at least make a copy or do a `mv` instead.

kkuehler: Yeah. If the user has any files in their public_html before the script is called I'll make a backup in their home folder. 

The wildcard character can match hidden files too. I just need to throw a `shopt -s dotglob` at the beginning of the script.

Baisang: Yeah `mv` would be good. Was going to say earlier that we want to create some sort of backup for the user.

makeservices/easywp
```diff
@@ -0,0 +1,38 @@
...
+php_defines 'username_here' "$USER"
+php_defines 'password_here' "$SQLPASS"
+php_defines 'localhost' 'mysql'
+php_defines 'utf8' 'utf8mb4'
+
+# Grab auth keys and salts from WP API
+sed -i '/put your unique phrase here/d' "$WPCONFIG"
+echo >> "$WPCONFIG"
+curl -sS -X GET https://api.wordpress.org/secret-key/1.1/salt/ >> "$WPCONFIG"
+
```
tahabi: you may also want to have some kind of output here giving the user confirmation that the script finished executing successfully and that their wordpress is now ready for use.

kkuehler:
@matthew-mcallister Can you review this again? Thanks.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+if [ -n "$(ls -A)" ]; then
+    TS=$(date +"%Y-%m-%d_%H:%M:%S")
+    BACKUP="$HOME/webroot-$TS"
+    mkdir -p "$BACKUP"
+    echo "Webroot not empty. Moving current files to $BACKUP..." 
+    mv ./* "$BACKUP"
+fi
+
+# Get SQL password
+echo "Resetting database..."
```
matthew-mcallister: I would change this to `Resetting database password...` so it's clear that only the password is reset.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+
+# Creates symlink to user's webroot. If the symlink exists and files
+# are present in webroot, they are backed up in the user's home directory.
+# A user's database is created or, if it already exists, the password 
+# is reset. WordPress is installed and the wp-config file is filled
+# out with the requisite information. Further configuration can be 
+# done by the user from a browser.  
+#
+# WARNING: This script should generally be run on a account without
+#          an existing website or database!
```
matthew-mcallister: I would be a little more explicit and summarize the potential danger: files get backed up, but the database password changes and the new WP install might try to migrate/clobber any existing database.

matthew-mcallister: If we expect users to run this, definitely make this a prompt before running. If only staff, it might be ok to leave it as a comment, but it still wouldn't hurt.

kkuehler: From talking with @Baisang, I thought that the user would click a button on the OCF website (https://www.ocf.berkeley.edu/account/commands/) to set up WordPress and it would run this script on tsunami as that user. Then they setup WordPress without having to touch a terminal. If that's the case there should be a warning prompt, but on the web interface. If not then I agree there should be a cli warning. That said, we could pipe 'y' into the script if it was going to be called in response to a button click.

matthew-mcallister: I'm honestly not sure if anyon ever visits the commands page anyways. I also think it's not a good idea to put this script up there. It can do all sorts of bad things if used improperly, and, if the script fails, whoever ran the command will have no recourse available from the website.

This script is probably best left for staff to run in the lab after setting up a new group account, or for somewhat experienced users who know how to SSH and can help themselves or get email help if things break. We don't want to encourage people to shoot themselves in the foot without knowing it.

chriskuehl: I agree with mattmcal. Additionally, I feel that we should eventually start to phase out "commands". It's a bizarre concept (run specific commands over SSH over HTTP)  when most of these thing should be first-class features.

Commands like `paper` are almost embarrassing -- we should just show the user's quota when they're logged in on the site. The same could be said for `makemysql` and `quota`. `makehttp` isn't even necessary anymore unless you mess things up, since we run it automatically for new accounts.

Baisang: I agree the commands section is not great, but just thought it would be a place to put something like this theoretically so users could do the WordPress install without touching a terminal. So if there were some tool that would let users automagically install WordPress for themselves, where would a better place be for it? Something like ocfweb just calling to ocflib?

Baisang: For now, I agree that it makes more sense to just have this run by staff or something. So the prompt Matt suggested would be a good idea.

matthew-mcallister: If we were to do something like that, and I'm not saying we should, it could indeed be a celery task called by ocfweb. However, I don't really like the sound of having a celery task that just shells out to a bash script, nor do I like the sound of rewriting this script in python, since it depends on `wp-cli` and `make{http,mysql}`, so I'm not even sure how this hypothetically code would be organized.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+
+# dos2unix
+sed -i 's/\r//' "$WPCONFIG"
+
+# Modify the config file with correct info
+php_defines 'database_name_here' "$USER"
+php_defines 'username_here' "$USER"
+php_defines 'password_here' "$SQLPASS"
+php_defines 'localhost' 'mysql'
+php_defines 'utf8' 'utf8mb4'
```
matthew-mcallister: It just occurred to me---couldn't you do `wp config set ...` here? Would it make the script simpler/easier to understand?

kkuehler: Oh rip... ```wp-config``` also grabs the salts too. Had no idea that existed. Yeah, that'll definitely simplify things.

makeservices/easywp
```diff
@@ -0,0 +1,67 @@
...
+
+set -euo pipefail
+shopt -s dotglob
+
+echo -e "\033[31mWARNING: This script should generally be run on a fresh account!\033[00m"
+echo -e "A user with an existing website or database will have their files 
+backed up, but their database password will change and the WordPress 
+install may clutter or try to migrate tables.\n"
+
+read -rp "Are you sure you want to continue? [Y/n] " idunno
```
matthew-mcallister: Nit: `n` should be capitalized since it's the default, not `y`.

kkuehler:
@matthew-mcallister Requested changes added. Also put in one more prompt if the user has files in webroot before doing the WP install.

makeservices/easywp
```diff
@@ -0,0 +1,67 @@
...
+
+# Create symlink to webroot
+makehttp
+cd "$webroot"
+
+# Check if user has files and let user decide what they want to do. 
+if [[ -n "$(ls -A)" ]]; then
+    echo
+    echo -e "There are currently files in webroot!\n"
+    read -rp "Do you want to back these files up? [Y/n/x] " whoknows
```
matthew-mcallister: Why even give the choice to nuke their files? Just notify the user that they will be backed up and ask if they want to continue.

makeservices/easywp
```diff
@@ -0,0 +1,67 @@
...
+# A user's database is created or, if it already exists, the password 
+# is reset. WordPress is installed and the wp-config file is filled
+# out with the requisite information. Further configuration can be 
+# done by the user from a browser.  
+#
+
+set -euo pipefail
+shopt -s dotglob
+
+echo -e "\033[31mWARNING: This script should generally be run on a fresh account!\033[00m"
```
matthew-mcallister: Nice colors. Maybe use yellow instead of red?

makeservices/easywp
```diff
@@ -0,0 +1,67 @@
...
+
+# Creates symlink to user's webroot. If the symlink exists and files
+# are present in webroot, they are backed up in the user's home directory.
+# A user's database is created or, if it already exists, the password 
+# is reset. WordPress is installed and the wp-config file is filled
+# out with the requisite information. Further configuration can be 
+# done by the user from a browser.  
+#
+
+set -euo pipefail
```
matthew-mcallister: I still think this script warrants the `-x`, **especially** the part where you back up the user's web dir. If you *really* want to make the script less noisy, I suggest you try a workaround. For example, you can do `set +x` inside a command group:
```bash
set -x
echo hello
(
    set +x
    echo world
)
```
There are alternatives; e.g. you could write a logging helper function and prefix your commands with it.

matthew-mcallister:
Protip: the first line in your git commit message is truncated after 50 chars. Subsequent lines are wrapped at 72 chars.

makeservices/easywp
kkuehler: On line 21 the regex ```^$``` will match someone just hitting enter (default) and there is a negation on the if-statement. Or are you asking me to make no the default option?

matthew-mcallister: Oh, I wasn't paying attention. Yes, make 'no' the default.

kkuehler:
@matthew-mcallister I know it's been a while, but wp cli was broken for a while. I made the changes you requested. Also I noticed the default permissions for wp-config.php was 644 and fixed that since a user's database password is in there.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+# done by the user from a browser.
+#
+
+set -euo pipefail
+shopt -s dotglob
+
+echo -e "\033[33mWARNING: This script should generally be run on a fresh account!\033[00m"
+echo -e "A user with an existing website or database will have their files
+backed up, but their database password will change and the WordPress
+install may clutter or try to migrate tables.\n"
```
matthew-mcallister: Minor point: to be completely accurate, this should read something like "WordPress may try to migrate and reuse existing tables". It certainly won't merely clutter them.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+
+# Create symlink to webroot
+makehttp
+cd "$webroot"
+
+# Check if user has files and let user decide what they want to do.
+if [[ -n "$(ls -A)" ]]; then
+    echo
+    echo -e "\033[33mThere are currently files in webroot!\033[00m\n"
+    read -rp "Do you want to continue? [y/N] " whoknows
```
matthew-mcallister: Should tell the user that their files will be backed up.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+
+    whoknows=${whoknows,,}
+    if [[ "$whoknows" =~ ^(yes|y) ]]; then
+        ts=$(date +"%Y-%m-%d_%H:%M:%S")
+        backup="$HOME/webroot-$ts"
```
matthew-mcallister: Minor suggestion: maybe use `public_html` as the prefix so it's more obvious what the directory contains to someone looking at it after the fact.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+        mv ./* "$backup"
+    else
+        echo "Ok, bye!"
+        exit 0
+    fi
+fi
+
+# Get SQL password
+echo "Resetting database password..."
+sqlpass=$(echo "yes" | makemysql | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
```
matthew-mcallister: It doesn't have to block this PR, but I feel like it would be better if `makemysql` had a `--noninteractive` option that would skip the confirmation prompt and spit out just the new password.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+if [[ -z "$sqlpass" ]]; then
+    echo -e "\033[31mError:\033[00m Could not retrieve database password. Run makemysql to see the issue."
+    exit 1
+fi
+echo "SQL set up and the password can be found in wp-config.php ..."
+
+# Install WordPress and create config file
+wp core download
+wp config create --dbname="$user" --dbuser="$user" --dbpass="$sqlpass" --dbhost=mysql --dbcharset=utf8mb4
+# Somehow 644 is the default...
```
matthew-mcallister: See https://core.trac.wordpress.org/ticket/37264. Heck, feel free to link to it in the comment.

makeservices/easywp
matthew-mcallister: I _still_ think this would be an improvement, but I won't make you do it. At lleast now it displays something for most of the stateful actions it takes.

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+
+echo -e "\033[33mWARNING: This script should generally be run on a fresh account!\033[00m"
+echo -e "A user with an existing website or database will have their files
+backed up, but their database password will change and the WordPress
+install may clutter or try to migrate tables.\n"
+
+read -rp "Are you sure you want to continue? [y/N] " idunno
+idunno=${idunno,,}
+if ! [[ "$idunno" =~ ^(yes|y) ]]; then
+    echo "Ok, bye!"
```
kpengboy: large

makeservices/easywp
```diff
@@ -0,0 +1,66 @@
...
+    echo -e "\033[33mThere are currently files in webroot!\033[00m\n"
+    read -rp "Do you want to continue? [y/N] " whoknows
+
+    whoknows=${whoknows,,}
+    if [[ "$whoknows" =~ ^(yes|y) ]]; then
+        ts=$(date +"%Y-%m-%d_%H:%M:%S")
+        backup="$HOME/webroot-$ts"
+        mkdir -p "$backup"
+        echo "Moving current files to $backup ..."
+        mv ./* "$backup"
```
kpengboy: Might want to remind that dotglob is in use?

PR #64, opened 2017-10-05
update ocf-tv to work without avahi

abizer:
also slightly better argparse and reorganize the script a bit

+1 to @matthew-mcallister for https://github.com/ocf/puppet/pull/195 for which this is the counterpart.

PR #65, opened 2017-10-09
Bump pre-commit hooks

jvperrin:
Some stuff in `ocf-tv` fails the checks too, but I don't want to mess with #64, so I'll leave that for now.


jvperrin:
I'll wait until #64 is merged for this one, this isn't urgent or anything.

PR #67, opened 2017-10-20
ssh-list: restrict filter for all hosts

kpengboy:
Let's not try to SSH into random dhcp-* hosts, or into blackhole, or into riptide-mgmt, or into lb (when we're already SSHing into whatever server lb actually happens to represent)

### DIFF
--- a/staff/sys/ssh-list
+++ b/staff/sys/ssh-list
@@ -36,7 +36,7 @@ def main(argv=None):
 
     args = parser.parse_args(argv)
     ldap_filter = {
-        'all': '(cn=*)',
+        'all': '(|(type=desktop)(type=server)(type=staffvm))',
         'desktop': '(type=desktop)',
         'server': '(|(type=server)(type=staffvm))',
     }[args.host_class]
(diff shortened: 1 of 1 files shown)


staff/sys/ssh-list
```diff
@@ -36,7 +36,7 @@ def main(argv=None):
 
     args = parser.parse_args(argv)
     ldap_filter = {
-        'all': '(cn=*)',
+        'all': '(|(type=desktop)(type=server)(type=staffvm))',
```
chriskuehl: should we blacklist the ones we don't want instead? i think we're more likely to add new types (e.g. we added staffvm recently) than we are to add new types we don't want to ssh into

matthew-mcallister: The currently used types are
```
desktop
dhcp
ipmi
printer
server
staffvm
switch
vip
wifi
```
We did add staffvm within the past four years, but we also added vip, so I think we've broken even. Unless we also added switch?

chriskuehl: True, it's probably about even. I'm just thinking it's worse not to get hosts you expected to get than to have extra hosts which don't actually break anything.

PR #70, opened 2017-11-05
Add updated timecard script

seanlobo:
I updated the [timecard script](https://github.com/ocf/labstats/blob/master/labstats/stats/timecard.py) from labstats, and @jvperrin mentioned it could go here.

### DIFF
--- a/staff/lab/timecard
+++ b/staff/lab/timecard
@@ -7,7 +7,7 @@ import matplotlib.pyplot as plt
 from ocflib.lab import stats
 
 
-UNIX_FILENAME_LIMIT = 255
+NAME_MAX = 255
 
 
 def generate_image(usage, users, dest, start, end):
@@ -95,19 +95,20 @@ def valid_date(s):
 if __name__ == '__main__':
     parser = argparse.ArgumentParser(description='Generate timecard graphs')
     parser.add_argument('users', type=str, nargs='+', help='User to generate graph for')
-    parser.add_argument('--dest', type=str, nargs='?', default='',
+    parser.add_argument('--dest', type=str, default='',
                         help='File to save graph in')
     parser.add_argument('--start', type=valid_date, default=None,
                         help='Start date for chart (in form YYYY-mm-dd; defaults to semester start')
     parser.add_argument('--end', type=valid_date, default=None,
-                        help='Start date for chart (in form YYYY-mm-dd; defaults to semester start')
+                        help='Start date for chart (in form YYYY-mm-dd; defaults to semester end')
     args = parser.parse_args()
+
     if args.dest:
         dest = args.dest
     else:
         # set name equal to a string of all users separated by underscores.
         # limit the string to be at max UNIX_FILENAME_LIMIT (-4 to account for '.png')
-        dest = '{}.png'.format('_'.join(sorted(args.users))[:UNIX_FILENAME_LIMIT - 4])
+        dest = '{}.png'.format('_'.join(sorted(args.users))[:NAME_MAX - 4])
 
     usage, start, end = get_usage_count(*args.users, start=args.start, end=args.end)
     generate_image(usage, args.users, dest, start, end)
--- a/staff/lab/timecard
+++ b/staff/lab/timecard
@@ -8,13 +8,17 @@ from ocflib.lab import stats
 
 
 def generate_image(usage, users, dest, start, end):
-    def h(h):
-        return 12 if h == 0 else (h if h <= 12 else h - 12)
+    def adjusted_hour(hour):
+        """Converts a 0-23 hour format into a 1-12 hour format"""
+        if hour == 0 or hour == 12:
+            return 12
+        else:
+            return hour % 12
 
-    def p(h):
-        return 'am' if h <= 11 else 'pm'
+    def pm_or_am(hour):
+        return 'am' if hour <= 11 else 'pm'
 
-    hours = ['{}{}'.format(h(hour), p(hour)) for hour in range(24)]
+    hours = ['{}{}'.format(adjusted_hour(hour), pm_or_am(hour)) for hour in range(24)]
     sums = [sum(usage[h]) for h in range(24)]
 
     fig = plt.figure()
@@ -29,18 +33,18 @@ def generate_image(usage, users, dest, start, end):
 
     plt.title('OCF Lab Timecard for {} ({}-{})'.format(
         ', '.join(users),
-        start.strftime('%m/%d/%Y'),
-        end.strftime('%m/%d/%Y')
+        start.strftime('%Y-%m-%d'),
+        end.strftime('%Y-%m-%d')
     ))
     plt.savefig(dest, bbox_inches='tight')
 
 
 def get_usage_count(*users, start=None, end=None):
-    dates = stats.semester_dates()
+    sem_start, sem_end = stats.semester_dates()
     if start is None:
-        start = datetime.datetime.combine(dates[0], datetime.time())
+        start = datetime.datetime.combine(sem_start, datetime.time())
     if end is None:
-        end = datetime.datetime.combine(dates[1], datetime.time())
+        end = datetime.datetime.combine(sem_end, datetime.time())
 
     step = datetime.timedelta(minutes=60)
     hours = collections.defaultdict(lambda: [0 for _ in range(7)])
@@ -60,9 +64,9 @@ def get_usage_count(*users, start=None, end=None):
     logins = cursor.fetchall()
 
     was_logged_in = dict()
-    for log in logins:
-        start_time = log['start'].replace(microsecond=0, second=0, minute=0)
-        end_time = log['end']
+    for session in logins:
+        start_time = session['start'].replace(microsecond=0, second=0, minute=0)
+        end_time = session['end']
         while start_time <= end_time:
             was_logged_in[start_time] = True
             start_time += step
@@ -81,7 +85,7 @@ def valid_date(s):
     try:
         return datetime.datetime.strptime(s, '%Y-%m-%d')
     except ValueError:
-        msg = "Not a valid date: '{0}'.".format(s)
+        msg = "Not a valid date: '{0}'. Use an isoformat string (YYYY-mm-dd)".format(s)
         raise argparse.ArgumentTypeError(msg)
 
 
@@ -95,7 +99,7 @@ if __name__ == '__main__':
     parser.add_argument('--end', type=valid_date, default=None,
                         help='Start date for chart (in form mm/dd/yyyy; defaults to semester start')
     args = parser.parse_args()
-    dest = args.dest if args.dest else ', '.join(args.users) + '.png'
+    dest = args.dest if args.dest else '{}.png'.format('_'.join(args.users)[:251])
 
     usage, start, end = get_usage_count(*args.users, start=args.start, end=args.end)
     generate_image(usage, args.users, dest, start, end)
--- a/staff/lab/timecard
+++ b/staff/lab/timecard
@@ -107,7 +107,7 @@ if __name__ == '__main__':
         dest = args.dest
     else:
         # set name equal to a string of all users separated by underscores.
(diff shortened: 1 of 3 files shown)


staff/lab/timecard
```diff
@@ -0,0 +1,101 @@
...
+        WHERE `user` IN %s
+            AND `start` >= %s
+            AND `end` <= %s
+            AND `duration` IS NOT NULL
+        ORDER BY start ASC"""
+    cursor.execute(query, (users, start, end))
+    logins = cursor.fetchall()
+
+    was_logged_in = dict()
+    for log in logins:
```
jvperrin: I'd probably rename `log` to something like `session` to be clearer what it refers to

staff/lab/timecard
```diff
@@ -0,0 +1,101 @@
...
+
+    return hours, start, end
+
+
+def valid_date(s):
+    # https://stackoverflow.com/a/25470943/6587177
+    try:
+        return datetime.datetime.strptime(s, '%Y-%m-%d')
+    except ValueError:
+        msg = "Not a valid date: '{0}'.".format(s)
```
jvperrin: I think putting something describing the correct format in here would be useful so if they get it wrong then it's clearer what the correct format they should be using it.

staff/lab/timecard
```diff
@@ -0,0 +1,101 @@
...
+        start.strftime('%m/%d/%Y'),
+        end.strftime('%m/%d/%Y')
+    ))
+    plt.savefig(dest, bbox_inches='tight')
+
+
+def get_usage_count(*users, start=None, end=None):
+    dates = stats.semester_dates()
+    if start is None:
+        start = datetime.datetime.combine(dates[0], datetime.time())
```
jvperrin: `datetime.datetime` here isn't great since it's a little confusing, but I'm not sure how you get around it besides importing is as a different name so that it doesn't mess up the other imports :(

Probably best to just leave it like this?

staff/lab/timecard
```diff
@@ -0,0 +1,101 @@
+#!/usr/bin/env python3
+import argparse
+import collections
+import datetime
+
+import matplotlib.pyplot as plt
```
kpengboy: @jvperrin should we also add `matplotlib` to `requirements-dev.txt`?

staff/lab/timecard
kpengboy: It should be possible to combine this loop with the one below, no?

staff/lab/timecard
kpengboy: Also a nice plus would be to make this work without requiring the `DISPLAY` variable set.

staff/lab/timecard
```diff
@@ -0,0 +1,101 @@
...
+    parser = argparse.ArgumentParser(description='Generate timecard graphs')
+    parser.add_argument('users', type=str, nargs='+', help='User to generate graph for')
+    parser.add_argument('dest', type=str, nargs='?', default='',
+                        help='File to save graph in')
+    parser.add_argument('--start', type=valid_date, default=None,
+                        help='Start date for chart (in form mm/dd/yyyy; defaults to semester start')
+    parser.add_argument('--end', type=valid_date, default=None,
+                        help='Start date for chart (in form mm/dd/yyyy; defaults to semester start')
+    args = parser.parse_args()
+    dest = args.dest if args.dest else ', '.join(args.users) + '.png'
```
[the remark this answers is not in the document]
kpengboy: Make it a required argument?

staff/lab/timecard
```diff
@@ -0,0 +1,105 @@
...
+
+def get_usage_count(*users, start=None, end=None):
+    sem_start, sem_end = stats.semester_dates()
+    if start is None:
+        start = datetime.datetime.combine(sem_start, datetime.time())
+    if end is None:
+        end = datetime.datetime.combine(sem_end, datetime.time())
+
+    step = datetime.timedelta(minutes=60)
+    hours = collections.defaultdict(lambda: [0 for _ in range(7)])
```
kpengboy: Could also write `[0] * 7`

staff/lab/timecard
```diff
@@ -0,0 +1,105 @@
...
+        WHERE `user` IN %s
+            AND `start` >= %s
+            AND `end` <= %s
+            AND `duration` IS NOT NULL
+        ORDER BY start ASC"""
+    cursor.execute(query, (users, start, end))
+    logins = cursor.fetchall()
+
+    was_logged_in = dict()
+    for session in logins:
```
kpengboy: It should be possible to combine this loop with the one below, no?

staff/lab/timecard
```diff
@@ -0,0 +1,105 @@
...
+    parser = argparse.ArgumentParser(description='Generate timecard graphs')
+    parser.add_argument('users', type=str, nargs='+', help='User to generate graph for')
+    parser.add_argument('dest', type=str, nargs='?', default='',
+                        help='File to save graph in')
+    parser.add_argument('--start', type=valid_date, default=None,
+                        help='Start date for chart (in form mm/dd/yyyy; defaults to semester start')
+    parser.add_argument('--end', type=valid_date, default=None,
+                        help='Start date for chart (in form mm/dd/yyyy; defaults to semester start')
+    args = parser.parse_args()
+    dest = args.dest if args.dest else '{}.png'.format('_'.join(args.users)[:251])
```
kpengboy: Where does `251` come from? Let's not use magic constants like this.

staff/lab/timecard
jvperrin: Yeah, that might be good, although ideally I think utils would have its own virtualenv at some point and be installed like ocflib is, as a `dh_virtualenv` package. But until then, adding `matplotlib` to `requirements-dev.txt` makes sense.

staff/lab/timecard
seanlobo: The reason I did this loop before the actual counting one is to account for concurrent logins. By first setting a boolean value for every hour a user is logged in from the it easily accounts for a user being logged in for a given hour twice, whereas I don't think it'd be as intuitive to do so with sessions. We'd need to somehow keep track of previous sessions and check if they overlap. It could be done though.

staff/lab/timecard
seanlobo: Ah, so it seems like the unix filename length is 255 characters, so I strip the prefix name to `251 + 4` from `.png` for a total of `255`. Should a comment be added for that?

kpengboy: Is there a built-in constant you can refer to?

seanlobo: I can't seem to find any from some google searches. It seems like it's platform dependent and on some OS's depends on the location of the file. Ok to just add a comment explaining where 251 comes from?

staff/lab/timecard
```diff
@@ -7,6 +7,9 @@ import matplotlib.pyplot as plt
 from ocflib.lab import stats
 
 
+UNIX_FILENAME_LIMIT = 255
```
kpengboy: Fwiw the standard name of this constant is NAME_MAX

staff/lab/timecard
matthew-mcallister: Isn't there an algorithm for combining overlapping intervals? Like, per-user, sort by start time, increment a counter when a session starts, and decrement when it ends?

staff/lab/timecard
```diff
@@ -0,0 +1,113 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Generate timecard graphs')
+    parser.add_argument('users', type=str, nargs='+', help='User to generate graph for')
+    parser.add_argument('--dest', type=str, nargs='?', default='',
+                        help='File to save graph in')
+    parser.add_argument('--start', type=valid_date, default=None,
+                        help='Start date for chart (in form YYYY-mm-dd; defaults to semester start')
+    parser.add_argument('--end', type=valid_date, default=None,
+                        help='Start date for chart (in form YYYY-mm-dd; defaults to semester start')
```
matthew-mcallister: You mean _end_ date?

staff/lab/timecard
```diff
@@ -0,0 +1,114 @@
...
+
+if __name__ == '__main__':
+    parser = argparse.ArgumentParser(description='Generate timecard graphs')
+    parser.add_argument('users', type=str, nargs='+', help='User to generate graph for')
+    parser.add_argument('--dest', type=str, default='',
+                        help='File to save graph in')
+    parser.add_argument('--start', type=valid_date, default=None,
+                        help='Start date for chart (in form YYYY-mm-dd; defaults to semester start')
+    parser.add_argument('--end', type=valid_date, default=None,
+                        help='Start date for chart (in form YYYY-mm-dd; defaults to semester end')
```
matthew-mcallister: This still says _start_ instead of _end_.

seanlobo: rip I'll fix it, thanks..

abizer:
I've been wondering if there was a way to do this entirely in SQL.

How long does this take to run?

seanlobo:
It takes about `1.1` seconds.

staff/lab/timecard
jvperrin: Ok, I added `matplotlib` to the `requirements-dev.txt`, and also fixed the output with no `$DISPLAY` set. Seems like a good idea for now, and then maybe use `dh_virtualenv` later.

seanlobo:
The latest commit modifies the graphs so that multiple users aren't grouped together into one query but instead stack on top of each other in various colors like in this picture

staff/lab/timecard
```diff
@@ -0,0 +1,139 @@
...
+    plt.ylabel('# Times Present')
+
+    # create a new mapping of user : length 24 array (frequencies for each hour)
+    users_daily_usage = dict()
+    for user, hours in usage.items():
+        users_daily_usage[user] = [sum(hours[h]) for h in range(24)]
+    stacked = [sum(x) for x in zip(*users_daily_usage.values())]
+    plt.ylim(0, max(stacked) * 1.15)  # set y limit to max cumulative usage
+
+    colors = ['b', 'y', 'g', 'r']
```
kpengboy: Why not just add all the standard colors

staff/lab/timecard
```diff
@@ -0,0 +1,139 @@
...
+        was_logged_in = dict()
+        for session in logins:
+            start_time = session['start'].replace(microsecond=0, second=0, minute=0)
+            end_time = session['end']
+            while start_time <= end_time:
+                was_logged_in[start_time] = True
+                start_time += step
+
+        time = start
+        while time <= end:
```
kpengboy: I will bitch one last time about this has bad performance, especially when # of hours logged in << # of hours in semester (which is always in practice)

staff/lab/timecard
```diff
@@ -0,0 +1,139 @@
...
+    fig.set_size_inches(12, 4)
+    plt.xlim(0, 24)
+    plt.xticks(range(24), hours, fontsize=8)
+    plt.xlabel('Hour')
+    plt.ylabel('# Times Present')
+
+    # create a new mapping of user : length 24 array (frequencies for each hour)
+    users_daily_usage = dict()
+    for user, hours in usage.items():
+        users_daily_usage[user] = [sum(hours[h]) for h in range(24)]
```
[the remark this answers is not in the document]
seanlobo: I'll change it to a dict comprehension, it was mainly like this so I could understand what I was writing at first.

staff/lab/timecard
matthew-mcallister: I thought I would see if it was actually faster to flatten overlapping sessions instead of using @seanlobo's algorithm. So I created this patch: https://github.com/matthew-mcallister/utils/commit/aeb37a65ccb47cf8dec1922ae2a5b5fbe662fa78. Then I generated this graph:
foo
Finally, I made the same performance measurement as @seanlobo:
```
Total query time: 0:00:01.150990
Total deduplication time: 0:00:00.000922
Total counting time: 0:00:00.004543
Total graphing time: 0:00:01.143586
```
Lo and behold, it's actually much slower than querying a dict once for each hour in the semester. That's Python for you!

seanlobo:
Is this good to be merged in?

PR #71, opened 2017-11-21
check: Don't silently strip bad input chars

kpengboy:
Also error out in case some variable was not defined.

### DIFF
--- a/acct/check
+++ b/acct/check
@@ -1,4 +1,5 @@
 #!/bin/sh
+set -u
 
 # Look up user for relevant account information.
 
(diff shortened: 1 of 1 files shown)


jvperrin:
This looks reasonable, but what was the motivation behind changing it? What inputs did you give that you thought should raise this error? I think some examples would make this clearer (and I think having those examples as comments in the script would help too)

chriskuehl:
more evidence that it's well past time to convert `check` to python

matthew-mcallister:
This script is kind of baffling. Why is it case-insensitive? When is the last time anyone used the `login_count` argument? Why don't we do the full `set -eux` at the top?

matthew-mcallister:
Also, I think the second commit can be merged onto master and doesn't belong in this PR.

kpengboy:
> Also, I think the second commit can be merged onto master and doesn't belong in this PR.

Sigh... That's what I get for assuming I was on master...

kpengboy:
I made these changes because `check`, unlike every other script, will strip out characters to make your input valid, rather than simply rejecting invalid input. I find this baffling behavior and think that `check` should do the latter like everyone else.

Other than also enabling `set -u` to catch uses of uninitialized variables, I have not attempted to make more changes.

> Why don't we do the full `set -eux` at the top?

We don't do `set -e` because IIRC it would break portions of this script without `-e` in mind. We don't do `-x` because we don't want all of the commands printed out, including all the sed's and grep's.

matthew-mcallister:
> We don't do -x because we don't want all of the commands printed out, including all the sed's and grep's.

OK, my bad, I meant `-o pipefail`. But the point is the quality of the script is questionable due to its age.

kpengboy:
Sure. I am not trying to do a general quality improvement of the script here, I am just trying to fix one particularly bizarre misfeature.

PR #78, opened 2018-02-02
Add "last for desktops" script

kpengboy:
This command is like (a subset of) the `last` command, except it
prints out logins across desktops, retrieved from the ocfstats
database.

### DIFF
--- a/staff/lab/last-lab
+++ b/staff/lab/last-lab
@@ -1,0 +1,62 @@
+#!/usr/bin/env python3
+import argparse
+import json
+import sys
+
+import ocflib.lab.stats
+from ocflib.infra.hosts import hostname_from_domain
+from tabulate import tabulate
+
+
+def make_fqdn(s):
+    """Argparse custom type that appends ".ocf.berkeley.edu" to short hostnames."""
+    if '.' not in s:
+        return s + '.ocf.berkeley.edu'
+    else:
+        return s
+
+
+def get_credentials():
+    """Return dictionary of ocfstats-ro creds, if we can read them."""
+    return json.load(open('/etc/ocfstats-ro.json', 'r'))
+
+
+def format_row(r):
+    """Given a row from the `session` table, formats it for printing."""
+    return (r['user'], hostname_from_domain(r['host']), r['start'],
+            'Still logged in' if r['end'] is None else r['end'])
+
+
+def main():
+    parser = argparse.ArgumentParser(description='Print out most recent desktop login records')
+    parser.add_argument('user', nargs='?', type=str,
+                        help='Show login records for the given user')
+    parser.add_argument('-n', '--limit', type=int, default=30,
+                        help='Only show this many entries')
+    parser.add_argument('-d', '--desktop', type=make_fqdn,
+                        help='Only show logins for this desktop')
+    args = parser.parse_args()
+
+    try:
+        creds = get_credentials()
+    except FileNotFoundError:
+        print('Could not find the file for ocfstats credentials.')
+        print('Are you running this on supernova?')
+        return 1
+
+    query = ('SELECT * FROM `session` WHERE '
+             + ' AND '.join((
+                 '`user` = %s' if args.user else '1',
+                 '`host` = %s' if args.desktop else '1',
+             ))
+             + ' ORDER BY `start` DESC LIMIT %s')
+    query_args = (*filter(None, (args.user, args.desktop)), args.limit)
+
+    with ocflib.lab.stats.get_connection(**creds) as c:
+        c.execute(query, query_args)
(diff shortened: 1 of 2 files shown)


staff/lab/last-lab
```diff
@@ -0,0 +1,62 @@
...
+    """Argparse custom type that appends ".ocf.berkeley.edu" to short hostnames."""
+    if '.' not in s:
+        return s + '.ocf.berkeley.edu'
+    else:
+        return s
+
+
+def get_credentials():
+    """Return dictionary of ocfstats-ro creds, if we can read them."""
+    return json.load(open('/etc/ocfstats-ro.json', 'r'))
```
jvperrin: I'm assuming this is coming in a puppet PR?

staff/lab/last-lab
```diff
@@ -0,0 +1,62 @@
...
+
+
+def format_row(r):
+    """Given a row from the `session` table, formats it for printing."""
+    return (r['user'], hostname_from_domain(r['host']), r['start'],
+            'Still logged in' if r['end'] is None else r['end'])
```
jvperrin: I think this would benefit from being split across more lines, something like:

```
return (
    r['user''],
    hostname_from_domain(['host']),
    r['start'],
    'Still logged in' if r['end'] is None else r['end'],
)
```

I think it just makes the number of items returned clearer and also makes it easier to see how each one is formatted.

staff/lab/last-lab
```diff
@@ -0,0 +1,62 @@
...
+import json
+import sys
+
+import ocflib.lab.stats
+from ocflib.infra.hosts import hostname_from_domain
+from tabulate import tabulate
+
+
+def hostname(s):
+    """Argparse custom type that appends ".ocf.berkeley.edu" to short hostnames."""
```
jvperrin: Pretty cool, I didn't know `argparse` could even do this.

staff/lab/last-lab
```diff
@@ -0,0 +1,62 @@
...
+        creds = get_credentials()
+    except FileNotFoundError:
+        print('Could not find the file for ocfstats credentials.')
+        print('Are you running this on supernova?')
+        return 1
+
+    query = ('SELECT * FROM `session` WHERE '
+             + ' AND '.join((
+                 '`user` = %s' if args.user else '1',
+                 '`host` = %s' if args.desktop else '1',
```
jvperrin: Why do it like this? I feel like just writing this more explicitly, even if some parts are repeated, would make this a whole lot clearer than executing a query with some extra integers in it that is hard to read.

staff/lab/last-lab
jvperrin: Also why not just make this a plain file with just a password instead of having to load it from JSON? Seems easier to me and less likely to break, and I don't see us changing the MySQL user all that often.

abizer:
this is the script I've been using for more or less this purpose for the last few months

https://i.fluffy.cc/MplFHQ9pR7Fczj9kLMbPnl70tqdXZpVm.html

staff/lab/last-lab
kpengboy: I largely copied this code from the `paper` script.

Sure, we could make it a plain text file. I guess that invites the question of whether we should make `paper` do the same though...

staff/lab/last-lab
kpengboy: I could do that, although it would make the code rather more verbose:

```python
query = 'SELECT * FROM `session`'
query_args = []
if args.user or args.desktop:
    query_conditions = []
    if args.user:
        query_conditions.append('`user` = %s')
        query_args.append(args.user)
    if args.desktop:
        query_conditions.append('`host` = %s')
        query_args.append(args.desktop)
    query += ' WHERE ' + ' AND '.join(query_conditions)
query += ' ORDER BY `start` DESC LIMIT %s'
query_args.append(args.limit)
```

kpengboy:
Wishlist: implement the `-p`, `-s`, `-t` options of `last`

staff/lab/last-lab
```diff
@@ -0,0 +1,92 @@
...
+    """Argparse custom type that appends ".ocf.berkeley.edu" to short hostnames."""
+    if '.' not in s:
+        return s + '.ocf.berkeley.edu'
+    else:
+        return s
+
+
+def build_query(args):
+    """Builds the MySQL query string with the given conditions."""
+    query = 'SELECT * FROM `session`'
```
abizer: `TIMESTAMPDIFF(start, end, minute)/60 as duration` is also a useful field to have, IMHO

kpengboy: I agree it could be useful. I'd prefer to implement it in `format_row` instead by subtracting `end` and `start` though. Topic for a future PR?

PR #79, opened 2018-02-17
last-lab: implement -s, -t, -p

kpengboy:
Implements the `-s`, `-t`, and `-p` options from the `last` command (although the semantics of `-s` are slightly different). This will make it easy to:

 - List which members are in the lab currently (`last-lab -p now`)
 - List which members were in the lab at 10:00 AM today (`last-lab -p 10:00`)
 - List which members have used a lab computer since midnight on February 1, 2018 (`last-lab -s 20180201`)
 - And more

### DIFF
--- a/staff/lab/last-lab
+++ b/staff/lab/last-lab
@@ -1,13 +1,11 @@
 #!/usr/bin/env python3
 import argparse
 import functools
-import re
 import sys
 import textwrap
 from datetime import date
 from datetime import datetime
 from datetime import time
-from datetime import timedelta
 
 import ocflib.lab.stats
 from ocflib.infra.hosts import hostname_from_domain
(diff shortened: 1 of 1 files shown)


staff/lab/last-lab
```diff
@@ -1,19 +1,82 @@
...
+    if '.' not in hostname:
+        return hostname + '.ocf.berkeley.edu'
     else:
-        return s
+        return hostname
+
+
+def parse_datetime(dtstr):
+    """Argparse custom type for parsing time expresssions."""
+    # TODO: consider switching to dateparser once that becomes feasible
```
jvperrin: Why is using dateparser not feasible? Do we need all these formats, or can we just have one standard one to make things easier?

staff/lab/last-lab
```diff
@@ -1,19 +1,82 @@
...
+from datetime import time
+from datetime import timedelta
 
 import ocflib.lab.stats
 from ocflib.infra.hosts import hostname_from_domain
 from tabulate import tabulate
 
 
-def make_fqdn(s):
+def make_fqdn(hostname):
```
jvperrin: Nice, I think this definitely helps with the readability of this function.

staff/lab/last-lab
chriskuehl: ++ on using some pre-existing library for this. we don't need to handle the exact same date formats that `last` accepts. this is an awful lot of unnecessary date-parsing logic.

staff/lab/last-lab
```diff
@@ -1,19 +1,82 @@
...
+    """Argparse custom type for parsing time expresssions."""
+    # TODO: consider switching to dateparser once that becomes feasible
+    FORMATS = ['%Y%m%d%H%M%S', '%Y-%m-%d %H:%M:%S', '%Y-%m-%d %H:%M',
+               '%Y-%m-%d']
+    FORMATS_TIME = ['%H:%M:%S', '%H:%M']
+    WORD_EXPR = {
+        'now': datetime.now(),
+        'yesterday': datetime.combine(date.today() - timedelta(days=1),
+                                      time()),
+        'today': datetime.combine(date.today(), time()),
```
chriskuehl: this indentation gave me cancer

staff/lab/last-lab
kpengboy: There's currently no Debian package for dateparser or dateutils, so unless we create a venv for utils or package them ourselves, it's not usable here.

I've tried parsedatetime too, but both that and dateparser give pathological results for some common date formats:

```
>>> import parsedatetime
>>> cal = parsedatetime.Calendar()
>>> cal.parse("20180223101845")  # This is one of `last`'s formats
(time.struct_time(tm_year=2018, tm_mon=2, tm_mday=23, tm_hour=22, tm_min=9, tm_sec=40, tm_wday=4, tm_yday=54, tm_isdst=0), 0)

>>> import dateparser
>>> dateparser.parse("20180223101845")
datetime.datetime(2033, 12, 12, 9, 45, 10)
```

What do you think? Should we limit it to one fixed format (plus "now" and "today")? Should we try to rope in one of the libraries anyway?

chriskuehl: https://packages.debian.org/stretch/python3-dateutil ?

just require yyyy-mm-dd imo, this should be plenty flexible enough

jvperrin: I think you might want to do something like `yyyy-mm-dd hh:mm` so that times can be allowed (we don't have second-based granularity anyway, so I'd just do to the minute), but I agree that just having one format would work fine as long as it's specified in the help output and is made clear when it doesn't work what the format should be. I don't think we really need to support all the possible formats if we don't have to.

kpengboy:
I've gotten rid of the relative date parsing logic which made up the bulk of that function. It's a lot shorter now. Let me know if you still think it's too complex.

PR #80, opened 2018-02-20
Updated VHost Disclaimer Script

Bagoum:
I've updated the vhost disclaimer checking script.

I added a script to calculate hashes from images, and I've also archived the current hosting images. There are some images not on ocfweb that are still being used, and I've kept the hashes for them from the old script.

I've added some bits to the regex to make them work for certain structures (I saw some places where the disclaimer was split by \<br> or \<li>).  I also added a check for the default broken site (see https://wrestling.berkeley.edu/). 

As a note, some images have carriage returns in them (windows shenanigans perhaps), so I've added multiple checks for the hash in the image-checking part.

I've also added a script that can check for offsite redirects. Get the banhammer ready!


abizer:
You'll need to run `pre-commit` on this in order for the build to pass (it's in the makefile). Also, I'd remove the date from the images directory. Could you also put the output of this somewhere in the staff directory?

staff/web/vhost/vhosting_check.py
```diff
@@ -43,14 +42,49 @@
     # OCF-Flag.png
     'ce48822dd63785f77d5826b6b215b18d',
     # lighter152x41.gif.png
-    'a9bfe3f918552692f5eddbd3c5446367'}
+    'a9bfe3f918552692f5eddbd3c5446367'} | \
```
chriskuehl: ? why union the two sets? can't you just add a comma there and continue the original set?

staff/web/vhost/get_new_hashes.py
```diff
@@ -0,0 +1,9 @@
+import hashlib
+import os
+
+dir = 'images_2-18-2018/'
+#Images from ocfweb/ocfweb/static/img/hosting-logos as of 2/18/18
+
+for f in os.listdir(dir):
+    with open(dir + f, 'rb') as fp:
+        print('#', f, "\n'" + hashlib.md5(fp.read()).hexdigest() + "',")
```
chriskuehl: Can we just calculate these in the script itself rather than run this and check in the hashes by hand? It will take approximately 0 seconds to calculate the hashes anyway.

staff/web/vhost/redirect_check.py
```diff
@@ -0,0 +1,48 @@
...
+                special_strings])
+
+def check_vhosting():
+    with open(out, 'w') as log, open(errors, 'w') as error_file:
+        vhosts = get_vhosts()
+        vhost_urls = []
+        for vhost_url in vhosts.keys():
+            if any(is_special(url) for url in {vhost_url} | set(vhosts[vhost_url]['aliases'])):
+                continue
+            vhost_urls.append('http://' + vhost_url)
```
chriskuehl: should this be https?

staff/web/vhost/vhosting_check.py
```diff
@@ -67,27 +101,40 @@ def check_vhosting():
...
             try:
-                site_opened = req.urlopen(site, timeout=10)
-                site_html = site_opened.readall().decode('utf-8')
-                counter = False
+                print("Opening", site)
+                site_html = r.get(site, timeout=10).text
+
+                isMissingDisc = False
+                if unavailable_site.search(site_html):
+                    raise Exception("Hosting down")
```
chriskuehl: should raise some more specific exception

staff/web/vhost/vhosting_check.py
```diff
@@ -67,27 +101,40 @@ def check_vhosting():
...
-                    if counter:
-                        m_b.writelines(site)
+                for img_url in img_urls:
+                    #I'm not exactly sure what the deal with \r is but this should work
+                    idata = r.get(img_url)
+                    img_hash_raw = hashlib.md5(idata.content).hexdigest()
+                    img_hash_unix = hashlib.md5(idata.content.replace(b'\r\n', b'\n')).hexdigest()
+                    img_hash_window = hashlib.md5(idata.content.replace(b'\n', b'\r\n')).hexdigest()
+                    if any([x in banner_hashes for x in [img_hash_raw, img_hash_unix, img_hash_window]]):
+                        break
```
chriskuehl: this is... bizarre. i assume this is for the SVGs?

staff/web/vhost/vhosting_check.py
```diff
@@ -43,14 +42,49 @@
...
 
 disclaimer_re = 'We\s*are\s*a\s*student\s*group\s*acting\s*independently\s*'\
-                'of\s*the\s*University\s*of\s*California.\s*We\s*take\s*full'\
+                'of\s*the\s*University\s*of\s*California.\s*(<[\s\S]*>)?\s*We\s*take\s*full'\
                 '\s*responsibility\s*for\s*our\s*organization\s*and\s*this'\
                 '\s*web\s*site.'
 disclaimer_pattern = re.compile(disclaimer_re)
-img_regex = re.compile("=\"?(\S+\.png|\S+\.gif|\S+\.jpg)")
+img_regex = re.compile("=\"?(\S+\.png|\S+\.gif|\S+\.jpg|\S+\.svg)")
+unavailable_site = re.compile("This\s*web\s*site\s*is\s*unavailable")
```
chriskuehl: can you check the status code instead? it should be a 503

staff/web/vhost/vhosting_check.py
ElDynamite: I found this issue when looking at PNGs. Certain versions had carriage returns in them and others didn't. This "workaround" should permit hash verification however the carriage returns are setup. (It may apply to SVGs too, I'm not sure)

staff/web/vhost/get_new_hashes.py
chriskuehl: Is it too slow or something? I'd expect it to take virtually no time at all to hash a dozen tiny images. That would be preferable to hard-coding hashes imo.

staff/web/vhost/redirect_check.py
```diff
@@ -0,0 +1,51 @@
...
+    return any([special_string in url_string for special_string in
+                special_strings])
+
+
+def check_vhosting():
+    with open(out, 'w') as log, open(errors, 'w') as error_file:
+        vhosts = get_vhosts()
+        vhost_urls = []
+        for vhost_url in vhosts.keys():
+            if any(is_special(url) for url in {vhost_url} | set(vhosts[vhost_url]['aliases'])):
```
kpengboy: FYI you can also write `{vhost_url, *vhosts[vhost_url]['aliases']}` since Python 3.5

staff/web/vhost/redirect_check.py
```diff
@@ -0,0 +1,51 @@
...
+        vhost_urls.sort()
+        baseIP = socket.gethostbyname('ocf.berkeley.edu')
+        for site in vhost_urls:
+            try:
+                print('Opening', site)
+                siter = r.get(site, timeout=10)
+                newIP = socket.gethostbyname(striphttp(siter.url))
+                if baseIP != newIP:
+                    print('\t', newIP)
+                    log.writelines(site + '\n')
```
kpengboy: `writelines` is for writing a list of strings. If you just want to write a string, use `write` instead. If you use `writelines`, then it will call `write` on each character individually, which is dumb.

Having said that, I would actually write this as `print(site, file=log)` instead.

staff/web/vhost/redirect_check.py
```diff
@@ -0,0 +1,51 @@
...
+
+            except Exception as e:
+                print('Error:', e)
+                error_file.writelines(site + '\n')
```
kpengboy: ibid (will stop commenting on this issue from here on)

staff/web/vhost/vhosting_check.py
```diff
@@ -43,14 +42,48 @@
...
+    '0276436d44d1a5e7cfc94b0f2d96907f',
+    # ocfbadge_silver8-b.png
+    'fc743ade8109379db4ddb2d4cf2723f3',
+    # ocfbadge_silver8.png
+    'a69a0929a9621638dd5b94161954fa5a'
+}
 
 disclaimer_re = 'We\s*are\s*a\s*student\s*group\s*acting\s*independently\s*'\
-                'of\s*the\s*University\s*of\s*California.\s*We\s*take\s*full'\
+                'of\s*the\s*University\s*of\s*California.\s*(<[\s\S]*>)?\s*We\s*take\s*full'\
```
kpengboy: Isn't `\s\S` equivalent to `.`? Or am I missing something?

And also, what's the motivation for this tag? A `<br>`?

staff/web/vhost/vhosting_check.py
```diff
@@ -67,27 +100,40 @@ def check_vhosting():
...
             try:
-                site_opened = req.urlopen(site, timeout=10)
-                site_html = site_opened.readall().decode('utf-8')
-                counter = False
+                print('Opening', site)
+                siter = r.get(site, timeout=10)
+                if siter.status_code == 503:
+                    raise Exception('503 Exception')
+                site_html = siter.text
+                isMissingDisc = False
```
kpengboy: Be careful to maintain consistency: this should be `is_missing_disc`

PR #82, opened 2018-07-05
check /opt/share/utils/bin/sorried for sorryfile

kkuehler:
It is no longer necessary to provide the full
path to a sorry file, though a full path will
still work. Also outputs all the sorry files in
the usage for ease of use.


staff/acct/sorry/sorry
```diff
@@ -32,6 +36,15 @@ if [ -z "$(getent passwd "$sorriedUser")" ]; then
 fi
 
 sorryFile=$2
+
+if [[ ! -f "$sorryFile" ]]; then
+    sorryFile="$REASONPATH$sorryFile"
```
abizer: for clarity I'd move the slash from $REASONPATH to here, i.e. $REASONPATH/$sorryFile

staff/acct/sorry/sorry
```diff
@@ -16,6 +17,9 @@ fi
 
 if [ -z "$1" ] || [ -z "$2" ]; then
     echo "Usage: $0 [user to be sorried] [sorry reason file]"
+    echo
+    echo "Sorry Reasons:"
+    find "$REASONPATH" -maxdepth 1 -mindepth 1 -type f -printf "   %f\n"
```
abizer: I'd also add e.g. "you can also use a custom sorry reason by passing in your own file" or something like that

staff/acct/sorry/sorry
```diff
@@ -32,6 +36,15 @@ if [ -z "$(getent passwd "$sorriedUser")" ]; then
 fi
 
 sorryFile=$2
+
+if [[ ! -f "$sorryFile" ]]; then
```
kpengboy: Would recommend using single brackets here in line with what we have in the rest of the script

jvperrin:
Why not just require this script be run on supernova and have an actual path for this that isn't variable? It seems to me having an actual log for this would be good, kinda like the note file (or put it in a database instead of a file).

kkuehler:
@kpengboy good now?

PR #83, opened 2018-09-24
Update to use Python 3.7 dh-virtualenv

ethanhs:
This is the foundation of the work updating utils to use dh-virtualenv for debian packaging.

Note: I updated python -> python2 shabangs while I was touching shabangs (I can split this out if you want)

Left to do:

- [x] test it works correctly
- [x] regen changelog/figure out versioning (maybe use Makefile like in approve?)
- [x] figure out what dependencies we need (@chriskuehl review of the Debian packaging would be appreciated)


debian/control
```diff
@@ -0,0 +1,11 @@
+Source: ocf-utils
+Section: python
+Priority: extra
+Maintainer: Open Computing Facility <help@ocf.berkeley.edu>
+Build-Depends: debhelper (>= 9), python, dh-virtualenv (>= 0.8)
+Standards-Version: 3.9.6
+
+Package: ocf-utils
+Architecture: any
+Depends: ${python:Depends}, ${misc:Depends}, ${shlibs:Depends}, virtualenv | python-virtualenv
```
kpengboy: Does this account for both Python 2 and Python 3?

requirements.txt
```diff
@@ -0,0 +1,5 @@
+dnspython
+matplotlib
+ocflib
```
kpengboy: I assume this downloads the latest `ocflib` from PyPI and bundles it?

ethanhs: Correct

setup.py
```diff
@@ -0,0 +1,31 @@
...
+shutil.rmtree(bin_dir, ignore_errors=True)
+
+sbin_dir = scripts_root / 'sbin'
+shutil.rmtree(sbin_dir, ignore_errors=True)
+
+shutil.copytree('./bin', bin_dir)
+shutil.copytree('./sbin', sbin_dir)
+
+setup(
+    name='approve',
```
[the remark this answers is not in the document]
ethanhs: D'oh, thanks

acct/check-dns
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python3
+#!/opt/venvs/ocf-utils/bin/python3.7
```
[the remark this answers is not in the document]
ethanhs: hm, that is an interesting idea. I'll try to add that tomorrow.

debian/control
ethanhs: I'm afraid I'm not sure what you mean. Account for 2 and 3 in what way?

jvperrin: It looks to me like due to the shebangs that python2 is used from the system whereas python3 is used from inside the packaged venv, which is fine. Definitely makes the packaging a bit easier anyway since we only have 2 python2 scripts and want to remove them soon.

kpengboy: But this package will include a dependency on both `python` and `python3` until such time as the last Python 2 script is migrated, no?

ethanhs: I guess it needs `python3.7` more specifically, in addition to `python`.

debian/control
```diff
@@ -0,0 +1,11 @@
+Source: ocf-utils
+Section: python
+Priority: extra
+Maintainer: Open Computing Facility <help@ocf.berkeley.edu>
+Build-Depends: debhelper (>= 9), python, dh-virtualenv (>= 0.8)
+Standards-Version: 3.9.6
+
+Package: ocf-utils
+Architecture: any
+Depends: ${shlibs:Depends}, virtualenv | python-virtualenv
```
kpengboy: Do we need an explicit dependency on python3?

ethanhs: It should depend on Python 2, and 3.7 at build time. There are still two scripts that use Python 2 at runtime, but otherwise we just need 3.7 at build time to make the virtualenv.

Makefile
```diff
@@ -3,11 +3,24 @@ test: venv
...
 
+.PHONY: builddeb
+builddeb: autoversion
+	dpkg-buildpackage -us -uc
+
+.PHONY: autoversion
+autoversion:
+	date +%Y.%m.%d.%H.%M-git`git rev-list -n1 HEAD | cut -b1-8` > .version
+	rm -f debian/changelog
+	DEBFULLNAME="Open Computing Facility" DEBEMAIL="help@ocf.berkeley.edu" VISUAL=true \
```
abizer: sm+packages@ocf.berkeley.edu perhaps?

debian/control
```diff
@@ -0,0 +1,11 @@
+Source: ocf-utils
+Section: python
+Priority: extra
+Maintainer: Open Computing Facility <help@ocf.berkeley.edu>
```
abizer: sm+packages again, or something like that, rather than sending it into RT

requirements-dev.txt
```diff
@@ -1,5 +1,2 @@
-dnspython
-matplotlib
-ocflib
+-r requirements.txt
```
abizer: til this works

Makefile
ethanhs: This was the same as in approve and the email we use for ocflib in the setup.py.

setup.py
```diff
@@ -0,0 +1,31 @@
...
+
+shutil.copytree('./bin', bin_dir)
+shutil.copytree('./sbin', sbin_dir)
+
+setup(
+    name='ocf-utils',
+    version=VERSION,
+    url='https://www.ocf.berkeley.edu/',
+    author='Open Computing Facility',
+    author_email='help@ocf.berkeley.edu',
```
abizer: sm+packages

setup.py
```diff
@@ -0,0 +1,31 @@
+import pathlib
```
abizer: is importing just the module a stylistic choice or something else? otherwise why not:

```python
from pathlib import Path
from shutil import rmtree
from shutil import copytree
```
just curious

vendor/venv-update
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
abizer: venv-update is python3-compatible, I thought? since this script is vendored, perhaps we should update it to the latest (3.1.1, from 3.0.0)?

acct/chsh
```diff
@@ -1,4 +1,4 @@
-#!/usr/bin/env python
+#!/usr/bin/env python2
```
abizer: this script is small enough that we can just write an LDIF to a tempfile and cat it into ldapmodify rather than sticking to python2 for the ldap library compatibility, if it makes a difference. we were able to do a proof of concept in about 20 minutes if you recall in the lab before the semester started.

Makefile
abizer: perhaps we should change it elsewhere as well, I don't have particularly strong opinions except that help@ seems inappropriate, being a customer service queue.

ethanhs: Yeah, it did seem weird. I'm okay changing it if we do so elsewhere.

acct/chsh
ethanhs: That is true, but I think that should be handled in a separate PR.

setup.py
ethanhs: Just stylistic, I don't feel strongly about this.

vendor/venv-update
ethanhs: Yes, that makes sense. I think though, that I would prefer to do that in a separate PR, I don't want to do too much in one change set.

abizer: sgtm

acct/chsh
abizer: sgtm

vendor/venv-update
jvperrin: The [latest still uses python2](https://github.com/Yelp/venv-update/blob/master/venv_update.py), but I don't think we really need to worry about this, since it's just a bootstrapping script meant to get dependencies and a virtualenv up and running, not one of our scripts, so it's always meant to use the system version of python 2 without any dependencies.

ethanhs: @jvperrin, does it? I don't see anything indicating it won't work with Python 3. Perhaps @asottile can say with certainty?

asottile: venv-update works fine on python 3

ethanhs: Good to know. Thanks :)

requirements.txt
jvperrin: Given this will now depend on `ocflib`, it's probably good to have the ocfweb build be an upstream build trigger to this repo so that this gets rebuilt when ocflib is updated.

Just for reference, this repo is a upstream trigger for the docker repo, since utils are included in each docker image, so building this repo (or ocflib) will cause quite a cascade.

Makefile
```diff
@@ -3,11 +3,24 @@ test: venv
 	venv/bin/pre-commit install
 	venv/bin/pre-commit run --all-files
 
+.PHONY: builddeb
+builddeb: autoversion
```
jvperrin: Following the [general package build pipeline](https://github.com/ocf/shared-pipeline/blob/205123a15fb74a9961d4a62f506f9b703bb275bf/vars/packagePipeline.groovy#L10), I think you'd also want to have a per distribution version of this like `package_%` (see [ocflib for inspiration](https://github.com/ocf/ocflib/blob/ae3bcade3290e921e41ab51c9ba5537c3e04efa5/Makefile#L28-L33)). Then have those output files into `dist_%`, and then the whole pipeline will handle the testing, build, and upload all automatically and the `Jenkinsfile` can just be like this or something (intentionally not including `jessie` since we should be fully moved off it in the next couple weeks now):

```
packagePipeline(
    dists: ['stretch', 'buster'],
)
```

acct/chsh
jvperrin: You'll want to update this once #87 is merged since then that'll need a python3 shebang too.

vendor/venv-update
jvperrin: You'll want to update this once #87 is merged since then that'll need a python3 shebang too.

requirements.txt
```diff
@@ -0,0 +1,5 @@
...
+setuptools
+tabulate
```
jvperrin: I think you'll also want:
- `mysqlclient`: https://github.com/ocf/utils/blob/f44b7ff9af0e57ea473041eb7d563453c57729ef/makeservices/makemysql-real#L27 Although maybe we can move this to something newer like `pymysql` or a celery task or something in the future.
- `ldap3`: https://github.com/ocf/utils/blob/f44b7ff9af0e57ea473041eb7d563453c57729ef/staff/acct/calnet-breakdowns#L6
- `progressbar`: https://github.com/ocf/utils/blob/f44b7ff9af0e57ea473041eb7d563453c57729ef/staff/acct/calnet-breakdowns#L11
- `pexpect`: https://github.com/ocf/utils/blob/f44b7ff9af0e57ea473041eb7d563453c57729ef/staff/acct/guess-passwords#L14
- `requests`: https://github.com/ocf/utils/blob/f44b7ff9af0e57ea473041eb7d563453c57729ef/staff/acct/guess-passwords#L15
- `lxml`: https://github.com/ocf/utils/blob/f44b7ff9af0e57ea473041eb7d563453c57729ef/staff/sys/apt-dater-ocf#L9
- `paramiko`: https://github.com/ocf/utils/blob/f44b7ff9af0e57ea473041eb7d563453c57729ef/staff/sys/makevm#L21

There's probably also some I'm missing here too.

Makefile
jvperrin: Yeah, this currently just spits out the package in the directory above `utils/`, which isn't great. Ideally we'd want this to be more self-contained and build in docker or something like that and then just split out the package files once done like builds in ocflib/approve/etc. do.

.gitignore
```diff
@@ -1,2 +1,3 @@
 *.py[co]
 /venv
+.version
```
jvperrin: I believe you'd want to ignore these files too that are generated by a build:

```
debian/.debhelper/
debian/changelog
debian/debhelper-build-stamp (maybe this should be checked in, all it has is "ocf-utils" as contents)
debian/files
debian/ocf-utils.debhelper.log
debian/ocf-utils.postinst.debhelper
debian/ocf-utils.substvars
debian/ocf-utils/
```

Makefile
```diff
@@ -3,11 +3,24 @@ test: venv
...
 venv: Makefile
 	vendor/venv-update \
-		venv= venv -ppython3 \
+		venv= venv -ppython3.7 \
 		install= -r requirements-dev.txt
 
 .PHONY: clean
 clean:
 	rm -rf venv
+	rm -rf debian/*.debhelper debian/*.log dist dist_*
```
jvperrin: nit: Combine these two lines?

requirements.txt
jvperrin: Ooh nice, thanks, that would be quite useful!

jvperrin: Hmmm, for some reason it only spat out `ocflib,requests,setuptools`, I think it's because it's only searching for files ending in `*.py` though: https://github.com/asottile/seed-isort-config/blob/55e34634d15a5a04fa92d5b1e963e989f4c96f12/seed_isort_config.py#L71

asottile: ah yeah indeed, if you have others scripts you'd have to list `--extra path/to/script1 --extra path/to/script2` -- didn't realize this had a bunch of extensionless oneoffs

jvperrin: Nice, that worked pretty well (although reasonably messy with all the `extra`s):
```
./venv/bin/seed-isort-config --extra acct/check-dns --extra makeservices/makehttp --extra makeservices/makemysql-real --extra net/pyrc --extra net/weather --extra printing/paper --extra staff/acct/calnet-breakdowns --extra staff/acct/chpass --extra staff/acct/guess-passwords --extra staff/acct/note --extra staff/acct/signat --extra staff/lab/check-toner --extra staff/lab/economode --extra staff/lab/lab-wakeup --extra staff/lab/last-lab --extra staff/lab/ocf-tv --extra staff/lab/timecard --extra staff/mail/extract-mail --extra staff/meetings/bod --extra staff/meetings/minutes --extra staff/sys/apt-dated --extra staff/sys/apt-dater-ocf --extra staff/sys/ldap-lint --extra staff/sys/makevm --extra staff/sys/migrate-vm --extra staff/sys/ocf-marathon --extra staff/sys/ocf-marathon --extra staff/web/bludgeon --application-directories acct:desktop:makeservices:net:printing:staff/acct:staff/lab:staff/mail:staff/meetings:staff/puppet:staff/ssl:staff/sys:staff/web:sys
```

```
[settings]
known_third_party = MySQLdb,dns,ldap3,lxml,matplotlib,ocflib,paramiko,pexpect,progressbar,requests,setuptools,six,tabulate,urllib3
```

Looks like I missed `six` before (I thought it was part of the stdlib), and `urllib3`, although that would be pulled in as a dependency of `requests` anyway.

debian/control
jvperrin: Well, the Python 2 requirement should be able to be removed now that the last two scripts have been migrated to Python 3.

requirements.txt
ethanhs: Thanks for digging into this! I think I have covered everything here.

ethanhs:
I am not convinced this is worth it. Due to the way Debian handles packages it would be a pain to manage this. Instead I think we should just wait till we move to Buster.

PR #84, opened 2018-09-27
updates to ocf-tv, change tunnel volume live

abizer:


staff/lab/ocf-tv
```diff
@@ -44,15 +44,15 @@ def wait_for_port(host, port, timeout=5, ssh_proc=None):
             raise TimeoutError('Timed out after {} seconds.'.format(timeout))
 
 
-def ssh_to_tv(host_port, remote_port):
+def tunnel_to_host(host_port, remote_port, remote_host):
```
kpengboy: Would call it `local_port` rather than `host_port` to avoid having multiple meanings of "host"

staff/lab/ocf-tv
```diff
@@ -62,14 +62,19 @@ def open_vnc(args):
...
+    proc = tunnel_to_host(port, 4713, args.host)
 
     try:
         wait_for_port('localhost', port, ssh_proc=proc)
         create_audio_tunnel(port)
-        input('Press enter to close the tunnel...')
+        while True:
+            vol = input('Enter new volume or nothing to close the tunnel: ')
+            if vol:
+                update_volume(check_volume(vol), args.host)
```
kpengboy: It would be better UX to not have the script bomb out if the user enters an invalid volume here. Consider catching the ArgumentTypeError and showing some kind of error message.

staff/lab/ocf-tv
```diff
@@ -135,6 +131,17 @@ def check_volume(val):
     return val
 
 
+def update_volume(amount, host):
+    cmd = ''
+
+    if amount:
+        cmd += 'pactl set-sink-volume {} {}%; '.format(TV_SINK, amount)
+
+    cmd += "pactl list sinks | awk '/Name: {}$/ {{ target = NR + 7 }}; NR == target {{ print $5 }}'".format(TV_SINK)
```
kpengboy: Don't forget to copy over the comment explaining how this line works.

staff/lab/ocf-tv
```diff
@@ -135,6 +131,17 @@ def check_volume(val):
     return val
 
 
+def update_volume(amount, host):
+    cmd = ''
```
kpengboy: So I take it we don't have to set PULSE_SERVER anymore?

staff/lab/ocf-tv
```diff
@@ -135,6 +131,17 @@ def check_volume(val):
...
+
+    if amount:
```
kpengboy: You may still wish to use `if amount is not None` here, lest funny things happen when a user tries to set the volume to 0 in interactive mode.

kpengboy: For that matter, it may be time to consider separating the functionality of setting volume from listing volume. `update_volume` is a misleading name if you're not updating the volume.

PR #85, opened 2018-10-05
add viyml, primarily for modifying staff hours

keur:


sys/viyml
```diff
@@ -0,0 +1,69 @@
...
+            target_file.write(content)
+            target_file.truncate()
+            break
+        except yaml.YAMLError:
+            pass
+
+        idunno = input('Syntax Error: Try again? [y/n] ')
+        idunno = idunno.strip().lower()
+        if idunno != 'y' and idunno != 'yes':
+            break
```
dkess: I'd put this question in a while loop, and make sure there's a clear "yes" or "no." Also, "no" should break the loop and "yes" should continue.

sys/viyml
```diff
@@ -0,0 +1,69 @@
...
+import errno
+import fcntl
+import signal
+from argparse import ArgumentParser
+from contextlib import contextmanager
+
+import yaml
+from ocflib.misc.shell import edit_file
+
+TIMEOUT = 2
```
dkess: Include the units of time in the variable name (`TIMEOUT_SECONDS`?)

sys/viyml
```diff
@@ -0,0 +1,69 @@
...
+
+        idunno = input('Syntax Error: Try again? [y/n] ')
+        idunno = idunno.strip().lower()
+        if idunno != 'y' and idunno != 'yes':
+            break
+
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument('path', nargs=1)
```
dkess: `nargs=1` isn't necessary, if you omit this option it won't return a list.

sys/viyml
```diff
@@ -0,0 +1,69 @@
...
+    with open(path, 'r+') as target_file:
+        with timeout(TIMEOUT):
+            try:
+                fcntl.flock(target_file.fileno(), fcntl.LOCK_EX)
+            except IOError as e:
+                if e.errno != errno.EINTR:
+                    raise e
+                print('Could not acquire lock. Someone else is modifying this file')
+                exit(1)
+        # we got a lock, now we can edit
```
dkess: Is the lock ever released?

keur: yeah, notice that `fcntl` takes a `fd`. after the `with` for the file ends the lock it is released implicitly. i could call `fcntl` to do this more clearly.

sys/viyml
keur: ah, thanks. will change this.

sys/viyml
```diff
@@ -0,0 +1,69 @@
...
+
+    with open(path, 'r+') as target_file:
+        with timeout(TIMEOUT):
+            try:
+                fcntl.flock(target_file.fileno(), fcntl.LOCK_EX)
+            except IOError as e:
+                if e.errno != errno.EINTR:
+                    raise e
+                print('Could not acquire lock. Someone else is modifying this file')
+                exit(1)
```
kpengboy: Use `sys.exit` instead of `exit`.

sys/viyml
```diff
@@ -0,0 +1,69 @@
...
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument('path', nargs=1)
+    args = parser.parse_args()
+
+    path = args.path[0]
+
+    with open(path, 'r+') as target_file:
+        with timeout(TIMEOUT):
```
kpengboy: Why the timeout? Unless we anticipate frequent situations in which the file is locked now, but will be unlocked within 2 seconds, I don't see this as being particularly useful. (Or is it NFS weirdness?)

sys/viyml
```diff
@@ -0,0 +1,69 @@
...
+            try:
+                fcntl.flock(target_file.fileno(), fcntl.LOCK_EX)
+            except IOError as e:
+                if e.errno != errno.EINTR:
+                    raise e
```
kpengboy: Since Python 3.3 it is possible to simply catch `InterruptedError` here.

sys/viyml
kpengboy: Eh, if it were me I would leave a comment around the `fcntl` explaining this instead.

sys/viyml
kpengboy: You had ought to provide help documentation for this command and for the argument.

sys/viyml
keur: It's more that if, say, I'm modifying the file, and you open the file, your terminal won't hang until you either `SIGINT` or I quit out from the file. I'm not attached to the timeout though

keur:
@kpengboy @dkess @ethanhs 

Resolved all the issues. Using the `LOCK_UN` flag in `fcntl` made this a lot cleaner. The timeout is no longer necessary. If we move forward with this, it will require one minor change to ocflib.

```
def get_staff_hours(content=None):                                                                                                                                                                                 
    if not content:                                                                                                                                                                                                
        staff_hours = _load_staff_hours()                                                                                                                                                                          
    else:                                                                                                                                                                                                          
        staff_hours = content  
    ...
    ...
```

sys/vistaff
```diff
@@ -0,0 +1,47 @@
...
+
+    content = None
+    with open(args.path, 'r') as target:
+        content = target.read()
+
+    with open(args.path, 'r+') as target:
+        try:
+            flock(target.fileno(), LOCK_EX | LOCK_NB)
+        except IOError:
+            print('Somebody else is modifying this file!')
```
dkess: Would it be possible to print who is modifying the file, and perhaps how long they've been editing it for? (ie if someone has been editing for 10 hours then there's probably a mistake)

sys/vistaff
```diff
@@ -0,0 +1,47 @@
...
+        while True:
+            content = edit_file(content)
+            try:
+                get_staff_hours(yaml.safe_load(content))
+                target.seek(0)
+                target.write(content)
+                target.truncate()
+                break
+            except (yaml.YAMLError, TypeError, KeyError):
+                pass
```
cg505: Why not just place the following 4 lines in the except block?

sys/vistaff
```diff
@@ -0,0 +1,47 @@
...
+            idunno = input('Syntax Error: Try again? [y/n] ')
```
cg505: best variable name

sys/vistaff
```diff
@@ -0,0 +1,47 @@
+#!/usr/bin/env python3
+from argparse import ArgumentParser
+from fcntl import flock
+from fcntl import LOCK_EX
+from fcntl import LOCK_NB
+from fcntl import LOCK_UN
+
+import yaml
+from ocflib.misc.shell import edit_file
+from staff_hours import get_staff_hours
```
cg505: This should be `ocflib.lab.staff_hours` right? Obviously only after the ocflib change is made.

sys/vistaff
```diff
@@ -0,0 +1,47 @@
...
+
+
+def main():
+    parser = ArgumentParser()
+    parser.add_argument('path')
```
cg505: Default to `~staff/lab_hours.yaml`?

sys/vistaff
dkess: I guess we can revisit this if it actually becomes a problem (which I suspect it will be) but for now I'm fine if this doesn't have it.

keur: Not pretty, but it'll work.
```
import subprocess
def get_access_time():
    pid = subprocess.check_output(['pgrep', '-o', 'easywp'])
    pid = pid.decode().splitlines()[0]
    time = subprocess.check_output(['ps', '-o', 'etime', '-o', 'ruser=RealUser', pid])
    pdata = time.decode().splitlines()[1].strip()
    time, user = pdata.split()
    return user, time

user, time = get_access_time()
print('{} has been modifying the file for {}'.format(user, time))
```

ethanhs:
@keur what was the status of this again?

keur:
@ethanhs hey, thanks for the poke. if it's a block i will fix it to be compatible with your changes later. if we need this merged i can prioritize it. lmk

ethanhs:
@keur oh yeah no rush. Just wanted to see where things were at. I'll probably work more on my PR next week. Thanks for the quick response :)

Issue #19, opened 2016-09-19
Convert python 2 scripts to python 3


kpengboy:
There are a few Python 2 scripts left; going with our broader migration to Python 3, we should convert these holdouts over to the current decade.

The specific scripts which aren't slated for removal according to my knowledge are:

```
./acct/chsh
./acct/update-email
./staff/acct/note
./staff/sys/apt-dated
./staff/web/bludgeon
./net/pyrc
```

chriskuehl:
+1

though "rewrite" is probably a bit strong of a word

update-email and chsh might be hard due to not being able to do gssapi for ldap (no python-ldap :()

matthew-mcallister:
We have an [ocflib API](https://github.com/ocf/ocflib/blob/master/ocflib/account/manage.py#L87) for this that works by shelling out to ldapmodify.

There will also be an account settings page in ocfweb in the near future and then no one will have to use update-email or chsh anyways, unless they want to.

chriskuehl:
that won't let you use gssapi though

chriskuehl:
it's not the end of the world, we can just ask people to enter their password again, but it's a kinda sucky regression

ideally we'd figure out how to get that gssapi support in ldap3 to work with our ldap/krb

matthew-mcallister:
Right, of course. At first I thought we could send it to create, but regular users can't submit celery tasks.

douglaslwong:
blockers: chsh, update-email

douglaslwong:
https://github.com/ocf/utils/pull/20 migrated all but chsh and update-email

douglaslwong:
reverting bludgeon for the time being

kpengboy:
bludgeon has been converted to python 3 in #69.

jvperrin:
This is done as of #87!

PR #87, opened 2018-11-07
Replace last python2 scripts with python3

jvperrin:
I originally rewrote `chsh` in bash, but decided rewriting it in python was better to be able to use some ocflib functions and generally be more maintainable.

This depends on the recently merged changes made in https://github.com/ocf/ocflib/pull/148 and in https://github.com/ocf/ocflib/commit/c2c5b9c5f17f139761027d95139209e3f566ee4f.

Testing done:
- `chsh`
  - With a valid shell and a valid and invalid kerberos ticket
  - With a invalid shell and a valid and invalid kerberos ticket
- `update-email`
  - With a invalid email already in my LDAP
  - With a valid email already, but no kerberos ticket
  - In auto mode with a valid/invalid email
  - With entering an invalid email to change to (including trying to change to an OCF email, which isn't allowed)

One thing to mention is that both scripts have a kind of race condition where it can check if you have a kerberos ticket, fetch a new one for you, and then you can destroy the ticket in another terminal or something before it gets to the part that actually tries to modify LDAP, but I don't think in practice that'll be a big deal, especially not for `chsh` since you'd have to do that quite quickly to get it to happen.

### DIFF
--- a/acct/chsh
+++ b/acct/chsh
@@ -6,32 +6,39 @@ import sys
 from ocflib.account.manage import modify_ldap_attributes
 from ocflib.misc.validators import VALID_LOGIN_SHELLS
 
-if len(sys.argv) < 2:
-    print('Usage: chsh [shell]')
-    sys.exit(1)
 
-new_shell = sys.argv[1]
+def main(args):
+    if len(args) < 2:
+        print('Usage: chsh [shell]')
+        return 1
 
-print('Changing shell to {}...'.format(new_shell))
+    new_shell = args[1]
 
-# Check if the user has a valid kerberos ticket
-has_kerberos = subprocess.run(('/usr/bin/klist', '-t')).returncode
+    print('Changing shell to {}...'.format(new_shell))
 
-if has_kerberos != 0:
-    # The user doesn't have a valid kerberos ticket already, so obtain a new
-    # ticket (this asks for a password if they don't already have a ticket, and
-    # then uses their password to obtain a new ticket)
-    subprocess.run('/usr/bin/kinit', check=True)
+    # Check if the user has a valid kerberos ticket
+    has_kerberos = subprocess.run(('/usr/bin/klist', '-t')).returncode
 
-# modify_ldap_attributes already does loginShell validation for us to make sure
-# it's a valid shell, so we don't have to do any extra here
-try:
-    modify_ldap_attributes(getpass.getuser(), {'loginShell': new_shell})
-except ValueError as e:
-    print(e, 'valid shells are: {}'.format(list(VALID_LOGIN_SHELLS)))
-    sys.exit(1)
+    if has_kerberos != 0:
+        # The user doesn't have a valid kerberos ticket already, so obtain a new
+        # ticket (this asks for a password if they don't already have a ticket, and
+        # then uses their password to obtain a new ticket)
+        subprocess.run('/usr/bin/kinit', check=True)
 
-print(
-    'Shell successfully changed, typically it takes a while for the '
-    'changes to propagate to all our machines'
-)
+    # modify_ldap_attributes already does loginShell validation for us to make sure
+    # it's a valid shell, so we don't have to do any extra here
+    try:
+        modify_ldap_attributes(getpass.getuser(), {'loginShell': new_shell})
+    except ValueError as e:
+        print(e, 'valid shells are: {}'.format(list(VALID_LOGIN_SHELLS)))
+        return 1
+
+    print(
+        'Shell successfully changed, typically it takes a while for the '
+        'changes to propagate to all our machines'
+    )
+    return 0
+
+
+if __name__ == '__main__':
+    exit(main(sys.argv))
--- a/acct/chsh
+++ b/acct/chsh
@@ -1,42 +1,37 @@
-#!/bin/bash
+#!/usr/bin/env python3
+import getpass
+import subprocess
+import sys
 
-if [ "$#" -ne 1 ]; then
-    echo 'Usage: chsh [shell]'
-    exit 1
-fi
+from ocflib.account.manage import modify_ldap_attributes
+from ocflib.misc.validators import VALID_LOGIN_SHELLS
 
-shell_choice=$1
+if len(sys.argv) < 2:
+    print('Usage: chsh [shell]')
+    sys.exit(1)
 
-ldif_template=$(cat <<EOL
-dn: uid=$(whoami),ou=people,dc=ocf,dc=berkeley,dc=edu
-changetype: modify
-replace: loginShell
-loginShell: $shell_choice
-EOL
+new_shell = sys.argv[1]
+
+print('Changing shell to {}...'.format(new_shell))
+
+# Check if the user has a valid kerberos ticket
+has_kerberos = subprocess.run(('/usr/bin/klist', '-t')).returncode
+
+if has_kerberos != 0:
+    # The user doesn't have a valid kerberos ticket already, so obtain a new
+    # ticket (this asks for a password if they don't already have a ticket, and
+    # then uses their password to obtain a new ticket)
+    subprocess.run('/usr/bin/kinit', check=True)
+
+# modify_ldap_attributes already does loginShell validation for us to make sure
+# it's a valid shell, so we don't have to do any extra here
+try:
+    modify_ldap_attributes(getpass.getuser(), {'loginShell': new_shell})
+except ValueError as e:
+    print(e, 'valid shells are: {}'.format(list(VALID_LOGIN_SHELLS)))
+    sys.exit(1)
+
+print(
+    'Shell successfully changed, typically it takes a while for the '
+    'changes to propagate to all our machines'
 )
-
-while read -r shell; do
-    # Strip any whitespace from the shell path
-    shell="${shell## }"
-    shell="${shell%% }"
-
-    # Check that the shell choice isn't commented out and is actually a valid
-    # listed shell
-    if [ "${shell:0:1}" != "#" ] && [ "$shell_choice" = "$shell" ]; then
-        echo "Changing shell to $shell"
-        /usr/bin/klist -t || /usr/bin/kinit
-
-        # Apply the change to LDAP and silence the (rather verbose) output
-        echo "$ldif_template" | ldapmodify -Q > /dev/null
-
-        echo 'Shell successfully changed, typically it takes a while for the' \
-             'changes to propagate to all our machines'
-
-        exit 0
-    fi
-done < /etc/shells
-
-echo "The shell you chose ($shell_choice) is not valid, please chose another"\
-     'shell'
-
-exit 1
--- a/acct/chsh
+++ b/acct/chsh
@@ -1,38 +1,42 @@
-#!/usr/bin/env python
-from __future__ import print_function
+#!/bin/bash
 
-import getpass
-import os
-import sys
+if [ "$#" -ne 1 ]; then
+    echo 'Usage: chsh [shell]'
+    exit 1
+fi
 
-import ldap
-import ldap.modlist as modlist
-import ldap.sasl as sasl
+shell_choice=$1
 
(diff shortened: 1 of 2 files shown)


jvperrin:
(oh, this closes #19, forgot to mention that)

acct/update-email
```diff
@@ -34,111 +26,104 @@ there is a problem with your account.
...
-        if valid_email(email):
+        if valid_non_ocf_email(email):
             sys.exit(0)
         else:
             print(AUTO_MESSAGE)
 
     print('Current Email Address: {}'.format(email or '(unset)'))
 
-    if valid_email(email) and input(Q_UPDATE_EMAIL) not in ('y', 'yes'):
+    if valid_non_ocf_email(email) and input(Q_UPDATE_EMAIL) not in ('y', 'yes'):
```
kpengboy:
```suggestion
    if valid_non_ocf_email(email) and input(Q_UPDATE_EMAIL) not in {'y', 'yes'}:
```

This is faster

acct/chsh
```diff
@@ -1,38 +1,37 @@
...
-print('The shell you chose is not valid, please chose another shell')
+    print('Usage: chsh [shell]')
+    sys.exit(1)
+
+new_shell = sys.argv[1]
+
+print('Changing shell to {}...'.format(new_shell))
+
+# Check if the user has a valid kerberos ticket
+has_kerberos = subprocess.run(('/usr/bin/klist', '-t')).returncode
```
abizer: perhaps use `subprocess.check_call`?

acct/update-email
abizer: smh ethanhs is not google scale

dkess: After running
```
python3 -m timeit "'y' not in {'y', 'yes'}"
python3 -m timeit "'y' not in ('y', 'yes')"
```

On `supernova`, I'm not actually convinced that either of these is faster than the other.

acct/chsh
ethanhs: I don't see a reason to. Also the Python docs say this about run:
>The recommended approach to invoking subprocesses is to use the run() function for all use cases it can handle.

acct/update-email
ethanhs: Agreed, I took a look and `set` doesn't implement `tp_as_sequence`, while of course `tuple` does. So theoretically tuple should be faster (and testing indicates this).

kpengboy: It actually depends. If you try running

```
python3 -m timeit "'n' not in {'y', 'yes'}"
python3 -m timeit "'n' not in ('y', 'yes')"
```

you might actually notice a difference.

In any case, no one is going to notice. Actually, a better argument might be semantic (think math notation).

acct/update-email
```diff
@@ -34,111 +26,104 @@ there is a problem with your account.
...
+    # something like ldap3.
+    output = subprocess.check_output(
+        ('ldapsearch', '-LLL', 'uid={}'.format(username), LDAP_MAIL_ATTR),
+        stderr=subprocess.DEVNULL,
+    ).decode('utf-8').split('\n')
 
+    mail_attr = list(filter(
+        lambda attr: attr.startswith('{}: '.format(LDAP_MAIL_ATTR)),
+        output
+    ))
```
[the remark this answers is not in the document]
kpengboy: Amended my comment

acct/update-email
ethanhs: >It actually depends.

Yeah, that is a good point. I should have been more careful in benchmarking claims.

>a better argument might be semantic (think math notation).

I understand where you are coming from, however Python is not math, and I think it is generally more Pythonic to use a tuple here. Furthermore, looking at sourcegraph, most uses of `x in <sequence>`  use a tuple, so for the sake of consistency I would argue a tuple is the best choice, unless there is a reason not to use one.

acct/chsh
```diff
@@ -1,38 +1,37 @@
...
 import getpass
-import os
+import subprocess
 import sys
 
-import ldap
-import ldap.modlist as modlist
-import ldap.sasl as sasl
+from ocflib.account.manage import modify_ldap_attributes
+from ocflib.misc.validators import VALID_LOGIN_SHELLS
```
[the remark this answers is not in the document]
jvperrin: Yeah, good idea, I'll do that.

acct/update-email
```diff
@@ -34,111 +26,104 @@ there is a problem with your account.
...
+def main(argv=None):
     parser = argparse.ArgumentParser(description='Update OCF email')
-    parser.add_argument('--auto', action='store_true',
-                        help='only prompt if email is missing, suppress errors')
+    parser.add_argument(
+        '--auto',
+        action='store_true',
+        help='only prompt if email is missing, suppress errors',
+    )
     args = parser.parse_args()
```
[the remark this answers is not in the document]
jvperrin: Why?

acct/update-email
jvperrin: Addressing the two separate points that have been brought up here:

1. Performance of a two element tuple/set absolutely **do not matter** here. Unless this is taking some outrageous amount of time (like more than half a second or something), then performance-motivated changes should not be motivating changes on this.

2. A semantic argument makes a whole lot more sense, but given this is Python and @ethanhs's argument on this, I think sticking with a tuple makes sense and also is clearer to read since it's clear that it's a tuple and not a hash or something that also uses `{` and `}`.

Thoughts? Personally, I don't really think it matters, and the amount of discussion here is way more than is needed for such a small issue/change.

acct/chsh
jvperrin: > check_call() and check_output() will raise CalledProcessError if the called process returns a non-zero return code.

(from https://docs.python.org/3.5/library/subprocess.html#exceptions)

Given we don't want an exception here and just want the return code, I think using `subprocess.run` here is better.

acct/update-email
ethanhs: I think my final opinion on the matter is the bikeshed should be blue.

acct/update-email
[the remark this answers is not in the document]
jvperrin: To follow up to my prior question, why would we use `argv` here when we can just not have to worry about manual argument parsing ourselves and have nice help strings provided by argparse?

acct/update-email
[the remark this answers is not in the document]
jvperrin: Yeah, this is definitely better, I'll use this instead of `filter`.

acct/update-email
```diff
@@ -1,31 +1,23 @@
...
-
-EMAIL_REGEX = re.compile('^[A-Z0-9._%+-]+@[A-Z0-9.-]+\.[A-Z]{2,4}$',
-                         re.IGNORECASE)
-
+LDAP_MAIL_ATTR = 'mail'
 Q_UPDATE_EMAIL = 'Would you like to update your email? [yN] '
 Q_NEW_EMAIL = 'New Email Address: '
-AUTO_MESSAGE = """
+
+AUTO_MESSAGE = '''
```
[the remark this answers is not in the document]
jvperrin: Meh, I don't really think this matters much and we aren't interpolating any variables into this or anything, but double quotes are also fine here.

acct/update-email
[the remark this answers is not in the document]
kpengboy: I was talking about the `argv` parameter of this function which you never use.

acct/update-email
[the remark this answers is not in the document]
kpengboy: Well, PEP 8 actually does say to use `"""` and not `'''`. In any case, I will not press the point further.

acct/chsh
```diff
@@ -6,32 +6,39 @@ import sys
...
-    sys.exit(1)
-
-print(
-    'Shell successfully changed, typically it takes a while for the '
-    'changes to propagate to all our machines'
-)
+
+def main(args):
+    if len(args) < 2:
+        print('Usage: chsh [shell]')
```
kpengboy: The `shell` argument is actually mandatory.

I guess this is a topic for a separate PR, but can we change this script to conform more closely to the actual `chsh`? (E.g. use `-s` to specify the shell, if no shell specified then prompt for one)

acct/chsh
```diff
@@ -6,32 +6,39 @@ import sys
...
+
+    print(
+        'Shell successfully changed, typically it takes a while for the '
+        'changes to propagate to all our machines'
+    )
+    return 0
+
+
+if __name__ == '__main__':
+    exit(main(sys.argv))
```
kpengboy: This should be `sys.exit` rather than `exit`

acct/update-email
```diff
@@ -34,111 +26,102 @@ there is a problem with your account.
...
 
+    mail_attr = [attr for attr in output if attr.startswith(LDAP_MAIL_ATTR + ': ')]
 
-def set_email(conn, dn, old_email, new_email):
-    """Updates email attribute in LDAP."""
-    old = {MAIL_ATTR: '*'}
-    new = {MAIL_ATTR: new_email}
+    if mail_attr:
+        # Strip the '{LDAP_MAIL_ATTR}: ' from the beginning of the string
+        return mail_attr[0][len(LDAP_MAIL_ATTR) + 2:].strip()
```
kpengboy: Do we care to check that there is only one mail attribute?

acct/update-email
```diff
@@ -34,111 +26,102 @@ there is a problem with your account.
...
+            print('The email you entered was invalid; please try again.\n')
         new_email = input(Q_NEW_EMAIL).strip()
 
-    set_email(conn, dn, email, new_email)
+    set_email(username, new_email)
     print('Updated email to {}'.format(new_email))
+
+
+if __name__ == '__main__':
+    exit(main())
```
kpengboy: ibid

acct/update-email
jvperrin: If there was, what would we do? Return both? Error out? It seems to me like just returning the first one is a pretty sane behavior.

I ended up modifying the LDAP config (stored in `cn=config cn={0}core`) and changed the `mail` attribute so that it can only have a single value (now it's just like `loginShell` in that behavior), and I think that essentially resolves this issue. I only found one account that had two `mail` attributes, and fixed that by using the one that was listed on their website as their contact address. I did a search using this command:

```
ldapsearch -LLL uid=* mail | pcregrep -B 2 -M "mail: .+\nmail:"
```

acct/chsh
jvperrin: Good catch, it should have been marked mandatory.

As for the interface, I think that would probably be better for users to use the more standard interface so that it's easier to select a shell (interactively), but I doubt it's a big deal. Maybe if some scripts or programs expect to use the standard interface for `chsh` or something, but even that seems like a pretty obscure use case. Ultimately, since we've had `chsh` work like this for ages and haven't had any complaints about it that I know of, it doesn't seem like changing it would really have much impact either way.

abizer: I remember being confused the first time around having chsh not have the semantics I expected from other machines because it was custom @ OCF. If we're going to override defaults like we are doing here it's probably a good idea to at least retain the interface.

jvperrin: Ok, I've done this in the latest commit, I agree that making our `chsh` as standard as possible (while still editing the `loginShell` in LDAP) is a good move.

acct/chsh
```diff
@@ -1,38 +1,60 @@
...
+    args = parser.parse_args()
+
+    username = getpass.getuser()
+    print('Changing the login shell for {}'.format(username))
+
+    # If no shell was provided, enter interactive mode and prompt for a shell
+    if args.shell:
+        new_shell = args.shell
+    else:
+        print('Enter the new value, or press ENTER to keep your current shell')
```
kpengboy: I would go with "Enter the new shell"

jvperrin: I went with this since it's literally what the actual `chsh` prints, but I agree that `Enter the new shell` would be a clearer/better message:

```
$ chsh
Password: 
Changing the login shell for jvperrin
Enter the new value, or press ENTER for the default
	Login Shell [/usr/bin/zsh]:
```

PR #97, opened 2019-02-28
easywp: Only can be run on tsunami

keur:


keur:
Sometimes I run this on supernova on accident and get confused.

PR #104, opened 2019-04-16
Use packaged kubernetes-deploy

fydai:
This uses the packaged kubernetes-deploy as in https://github.com/ocf/kubernetes-deploy-deb instead of the given Dockerfile. kubernetes-deploy is in ocf::extrapackages, so it should be installed whenever kubectl is installed.

### DIFF
--- a/staff/sys/ocf-kubernetes-deploy
+++ b/staff/sys/ocf-kubernetes-deploy
@@ -9,8 +9,6 @@ import os
 import subprocess
 import sys
 
-KUBERNETES_DEPLOY_DOCKER = 'docker.ocf.berkeley.edu/kubernetes-deploy:latest'
-
 
 def get_current_context():
     return subprocess.check_output(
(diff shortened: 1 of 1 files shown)


staff/sys/ocf-kubernetes-deploy
```diff
@@ -67,20 +65,15 @@ def main():
...
-        '-t', KUBERNETES_DEPLOY_DOCKER,
-        namespace_name,
-        args.kube_context,
-        '--bindings=' + json.dumps(bindings),
-    ]).check_returncode()
+    subprocess.run(['kubernetes-deploy',
+                    namespace_name,
+                    args.kube_context,
+                    '--bindings=' + json.dumps(bindings),
+                    '--template-dir=' + args.dir]) .check_returncode()
```
dkess: Keep the indentation style from before

PR #105, opened 2019-04-26
Escape filter chars in guess-passwords

dkess:


kpengboy:
Honestly, how many other LDAP injection attacks are we vulnerable to...

dkess:
I think I caught all of them with the recent wave of PRs, it doesn't seem like a huge deal. This one especially isn't a huge deal.

PR #106, opened 2019-05-05
a small script to add users to ldap groups

abizer:
doing `kinit abizer/admin ldapvi cn=ocfhpc` or whatever isn't hard, but this is slightly nicer

### DIFF
--- a/staff/acct/add-to-group
+++ b/staff/acct/add-to-group
@@ -1,14 +1,33 @@
 #!/bin/bash
 set -euo pipefail
 
+if [ "$#" -ne 2 ]; then
+    echo "usage: $0 username groupname"
+    exit 1
+fi
+
 NEW_USER="$1"
 LDAP_GROUP="$2"
 
-ldapmodify <<EOF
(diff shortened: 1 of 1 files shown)


abizer:
hmm, yeah. I should also check `klist`

redplanks:
How does it fail if you supply an invalid group and/or user?

dkess:
IMO we should have a confirmation dialog ("are you sure?") since this has the potential to do dangerous stuff.

Otherwise lgtm, we should perhaps also modify necessary parts of our docs to use this command.

cg505:
@dkess++, maybe we should print the change before actually sending it to ldapmodify

staff/acct/add-to-group
```diff
@@ -1,14 +1,33 @@
 #!/bin/bash
 set -euo pipefail
 
+if [ "$#" -ne 2 ]; then
+    ehco "usage: $0 username groupname"
```
dkess: typo

PR #109, opened 2019-05-30
Updated kubernetes deploy

fydai:


staff/sys/ocf-kubernetes-deploy
```diff
@@ -61,20 +76,30 @@ def main():
...
-        'version': args.appversion,
-    }
 
-    # kubernetes-deploy requires that this environmental variable be set, and passes
-    # it to templates as the `current_sha` variable. We don't ever use it.
-    os.environ['REVISION'] = 'unused'
+    # Created with 600 perms
+    bindings_file = tempfile.NamedTemporaryFile(suffix='.yaml')
+
+    version_binding = 'version: ' + args.appversion + '\n'
```
dkess: This is not a good way to write YAML. You should keep everything in a `bindings` dict, and write it all to a temporary JSON file at the end.

staff/sys/ocf-kubernetes-deploy
```diff
@@ -61,20 +76,30 @@ def main():
...
+    bindings_file.write(version_binding.encode())
+
+    if os.path.isfile(args.secrets):
+        with open(args.secrets) as secrets_file:
+            bindings_file.write(secrets_file.read().encode())
+    else:
+        for secrets_path in os.listdir(args.secrets):
```
dkess: Instead of looping through all secrets and concatenating them, we should pick a single secrets file based on the `appname`. Otherwise how will potential key collisions be handled?

staff/sys/ocf-kubernetes-deploy
```diff
@@ -61,20 +76,30 @@ def main():
...
     # Bindings passed into the kubernetes-deploy templates
-    bindings = {
-        'version': args.appversion,
-    }
 
-    # kubernetes-deploy requires that this environmental variable be set, and passes
-    # it to templates as the `current_sha` variable. We don't ever use it.
-    os.environ['REVISION'] = 'unused'
+    # Created with 600 perms
+    bindings_file = tempfile.NamedTemporaryFile(suffix='.yaml')
```
dkess: You should use a `with` statement here, see https://github.com/ocf/utils/blob/352f6eb68434d73cf21acd80dee81dad4543abdc/staff/sys/ssh-list#L51-L56

staff/sys/ocf-kubernetes-deploy
```diff
@@ -36,13 +44,20 @@ def main():
              "app-{appname}".
         ''',
     )
+
+    # This argument is optional because some Kubernetes services
+    # (sourcegraph) don't need us to build our own Docker images
     parser.add_argument(
         'appversion',
+        nargs='?',
+        default='',
```
dkess: Have you tested the case where this isn't passed in? I believe `kubernetes-deploy` requires that the `version` variable is passed as a binding.

fydai: I have tested this, kubernetes deploy doesn't require a version binding. That was the REVISION binding and is no longer mandatory.

dkess: Oh cool that's good to know

PR #110, opened 2019-06-01
Don't fail when secrets missing

fydai:


dkess:
We still shouldn't be failing silently here. As I see it, there are three cases:

* User specifies secrets, and they don't exist. Fail loudly
* User specifies secrets and they do exist. Carry on
* User doesn't specify secrets at all. Do nothing

This PR will make the first case fail silently, which is bad for debugging.

fydai:
What makes this a bit more complicated is there are two other cases:
- Jenkins looks for a secrets file for the given app, and it doesn't exist. Carry on.
- Jenkins looks for a secrets file for the given app, and it does exist. Carry on.

dkess:
Hmmm, I guess it'll fail anyways if the templates are using variables that aren't defined. Maybe print a message saying "no secrets found" or something, to make the log clearer?

fydai:
Ninja'd by 3 seconds :P

PR #111, opened 2019-06-06
makeservices: add easydoku script

dkess:
Structured similarly to the easywp script, this installs DokuWiki. It includes a prompt for subfolder, since some groups might want this at a separate URL.

Can someone please test this? It wfm but I may have missed something important.


makeservices/easydoku
```diff
@@ -0,0 +1,68 @@
...
+read -r subfolder
+
+read -rp "Are you sure you want to continue? [y/N] " idunno
+idunno=${idunno,,}
+if ! [[ "$idunno" =~ ^(yes|y) ]]; then
+    echo "Ok, bye!"
+    exit 0
+fi
+
+webroot=$HOME/public_html/"$subfolder"
```
kpengboy:
```suggestion
webroot="$HOME/public_html/$subfolder"
```

makeservices/easydoku
```diff
@@ -0,0 +1,68 @@
...
+# Check if user has files and let user decide what they want to do.
+if [[ -n "$(ls -A)" ]]; then
+    echo
+    echo -e "\033[33mThere are currently files in webroot! These will be backed up.\033[00m\n"
+    read -rp "Do you want to continue? [y/N] " whoknows
+
+    whoknows=${whoknows,,}
+    if [[ "$whoknows" =~ ^(yes|y) ]]; then
+        ts=$(date +"%Y-%m-%d_%H:%M:%S")
+        backup="$HOME/public_html-$ts"
```
kpengboy: What should be the correct behavior if `webroot` is not the `public_html` directory?

dkess: I'll change this to `public_html-doku-$ts`. The behavior is still correct (since we `cd` into the directory Doku will be installed in.

dkess:
Made the changes. I don't think these failure cases are a huge deal, but if this gets used a lot and we see users confused by it, we should revisit.

makeservices/easydoku
```diff
@@ -0,0 +1,68 @@
...
+echo "Downloading and installing DokuWiki"
+curl -L "https://download.dokuwiki.org/src/dokuwiki/dokuwiki-stable.tgz" \
+    | tar -xz --strip-components=1
+
+echo "Downloading and installing the Upgrade plugin"
+mkdir lib/plugins/upgrade
+curl -L "https://github.com/splitbrain/dokuwiki-plugin-upgrade/tarball/master" \
+    | tar -xz -C lib/plugins/upgrade --strip-components=1
+
+echo "DokuWiki install successful. Go to https://www.ocf.berkeley.edu/~$user/$subfolder to finish the setup process."
```
cg505:
```suggestion
echo "DokuWiki install successful. Go to https://www.ocf.berkeley.edu/~$user/$subfolder/install.php to finish the setup process."
```

I needed this otherwise it just dropped me at a non-existent home page.

cg505:
```suggestion
echo "DokuWiki install successful. Go to https://www.ocf.berkeley.edu/~$user/$subfolder/install.php to finish the setup process."
```

I needed this otherwise it just dropped me at a non-existent home page.

dkess:
DokuWiki will be packaged in Buster, which means eventually we might not need to download from their website, with the added bonus that it'll be updated and managed by us, assuming we can get it to work with our Apache userdir setup.

ethanhs:
It looks to me like this probably could just be merged? I guess we could wait for bullseye, but it may be a while before tsunami is on bullseye...

dkess:
I'm probably OK with this being merged today. The OCF doesn't really have a precedent of using Debian packages for vhost software (though in theory we could do this for wordpress).

Anyone want to test it quickly one last time?

ethanhs:
Yeah I can test it in a little while.

ethanhs (approved):
This looks great!

ethanhs:
Also for some reason downloading docuwiki was at 100Kb/s, so we may want to consider mirroring it if possible.

PR #112, opened 2019-06-07
makeservices: add makexmpp script

dkess:
This lets users create their own XMPP accounts. The script might seem daunting but it's actually not as complicated as it seems. It connects as the XMPP admin and attempts to change the user's password. If that errors, it then tries to create the account. This relies on sudo rules in https://github.com/ocf/puppet/pull/659.


makeservices/makexmpp-real
```diff
@@ -0,0 +1,250 @@
...
+        You can always re-run this command to reset the password if you lose it.
+
+        If you are ready to continue, type 'yes'.
+        Typing anything other than yes will abort this script.
+        """
+    ))
+    return input('Continue? ') == 'yes'
+
+
+class XMPPUserPassword(sleekxmpp.ClientXMPP):
```
kpengboy:
```suggestion
class XMPPUserPasswordClient(sleekxmpp.ClientXMPP):
```

makeservices/makexmpp-real
```diff
@@ -0,0 +1,250 @@
...
+            username,
+            newpassword,
+        )
+        xmpp.register_plugin('xep_0133')  # Service Administration
+
+        # Connect to the XMPP server and start processing XMPP stanzas.
+        if xmpp.connect(reattempt=False):
+            xmpp.process(block=True)
+        else:
+            raise Exception('Unable to connect to XMPP server.')
```
kpengboy: Can we use a more specific subclass of Exception?

makeservices/makexmpp-real
```diff
@@ -0,0 +1,250 @@
...
+JID_DOMAIN = 'ocf.berkeley.edu'
+
+PW_LENGTH = 24
+
+
+def read_config():
+    """Fetches the server admin JID and password from the config in
+    '/opt/share/makeservices'."""
+    conf = ConfigParser()
+    conf.readfp(open(CONF_FILE))
```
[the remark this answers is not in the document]
dkess: This function was used in `easywp` but looks like it's been replaced. I'll change this to use `read`.

makeservices/makexmpp-real
```diff
@@ -0,0 +1,250 @@
...
+        provide any additional data.
+        """
+
+        def command_error(iq, session):
+            self['xep_0050'].terminate_command(session)
+            self.disconnect()
+
+            condition = iq['error']['condition']
+            errtext = iq['error']['text']
+            raise Exception('{}: {}'.format(condition, errtext))
```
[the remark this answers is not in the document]
dkess: I don't think this is a huge deal, since this is only seen by the emailed problem report in the `root@` mailing list. We'll see the line number which is more important anyways.

dkess:
Thanks for the feedback, I've made most of the suggested changes. Not opposed to colorized output but I'd rather do it in a separate commit, and it's not a huge priority IMO.

makeservices/makexmpp-real
```diff
@@ -0,0 +1,250 @@
+#!/usr/bin/env python3
+"""
+Creates an XMPP account with the same name as the username of the user who
+runs this program.
+
+To prevent the user from taking control of this program and e.g., causing the
+program to dump core and reveal the Prosody admin credentials, this program
+should be run under an unprivileged account.
```
kpengboy:
```suggestion
should be run under a separate account from that of the user.
```

PR #114, opened 2019-06-15
Add support for adjustments

glee-:
Depends on ocf/ocflib#182

>Allows staff members to adjust users' paper quotas in a more flexible manner.
>We add the following functionality to the `paper` script:
>`forward`:  increases a user's daily quota for the day only
>
>This is intended to be used to give users some degree of freedom when printing documents.
>
>For example, if the daily print limit was set at `10` pages and a user wanted to print `12` pages, this would not be possible with the current way things are configured. `forward` provides the way to increase the daily print limit without increasing the amount of pages a user can print in a semester.

### DIFF
--- a/printing/paper
+++ b/printing/paper
@@ -11,15 +11,15 @@ from ocflib.misc.shell import green
 from ocflib.misc.shell import red
 from ocflib.misc.shell import yellow
 from ocflib.misc.whoami import current_user
-from ocflib.printing.quota import add_refund
+from ocflib.printing.quota import add_adjustment
 from ocflib.printing.quota import get_connection
 from ocflib.printing.quota import get_quota
-from ocflib.printing.quota import Refund
+from ocflib.printing.quota import Payload
 
 
 def staff_creds():
     """Return dictionary of staff creds, if we can read them."""
-    return json.load(open('/etc/ocfprinting.json'))
+    return json.load(open('/etc/ocfprinting-dev.json'))
 
 
 def view(args):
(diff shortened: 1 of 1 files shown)


printing/paper
```diff
@@ -11,15 +11,15 @@ from ocflib.misc.shell import green
...
 from ocflib.printing.quota import get_connection
 from ocflib.printing.quota import get_quota
-from ocflib.printing.quota import Refund
+from ocflib.printing.quota import Payload
 
 
 def staff_creds():
     """Return dictionary of staff creds, if we can read them."""
-    return json.load(open('/etc/ocfprinting.json'))
+    return json.load(open('/etc/ocfprinting-dev.json'))
```
abizer: fix this for prod

printing/paper
```diff
@@ -87,6 +88,11 @@ def main(argv=None):
     parser_refund.add_argument('--reason', '-r', required=True, type=str)
     parser_refund.add_argument('user', type=str)
 
+    parser_forward = subparsers.add_parser('forward', help="allow a user to print above today's quota")
+    parser_forward.add_argument('--pages', '-p', required=True, type=int)
+    parser_forward.add_argument('--reason', '-r', required=True, type=str)
```
MrMinos: Forward probably in itself is the reason, I don't know if this needs to be a required field. What do you think?

printing/paper
```diff
@@ -96,7 +102,10 @@ def main(argv=None):
         print(bold(red("The user {} doesn't exist.".format(args.user))))
         return 1
 
-    return commands[args.command](args)
+    if args.command == 'refund' or args.command == 'forward':
```
dkess: Instead of an if statement here, which makes the subcommand logic harder to understand, you should put the same if statement in the `adjust` command (which has access to `args` and therefore `args.command`).

printing/paper
dkess: I think this is still fine, and it's a database constraint anyways

printing/paper
glee-: good catch

printing/paper
glee-: Despite the reason being clear, I think this would be important to include due to the human aspect of it. A person who is issuing the forward request will need to think twice before actually increasing the daily quota for a particular user.

printing/paper
```diff
@@ -43,36 +43,37 @@ def view(args):
...
 
 
-def refund(args):
-    refund = Refund(
+def adjust(args, type):
+    payload = Payload(
         user=args.user,
         time=datetime.now(),
         pages=args.pages,
+        action=type,
```
TonyLianLong: You can just leave `refund` as a method with 1 argument, but use `args.command` for the `action` parameter instead of adding parameter `type` (This is probably what @dkess says and in this case my comment can serve as a clarification).

printing/paper
```diff
@@ -43,36 +43,37 @@ def view(args):
...
         staffer=current_user(),
         reason=args.reason,
     )
-    prompt = bold('Refund {} pages to {}? [yN] '.format(refund.pages, refund.user))
+
+    stringtype = 'Refund' if type == 'refund' else 'Forward'
+    prompt = bold(stringtype + ' {} pages to {}? [yN] '.format(payload.pages, payload.user))
```
cg505: If we're already using format here, just use it consistently.
```suggestion
    prompt = bold('{} {} pages to {}? [yN] '.format(stringtype, payload.pages, payload.user))
```

PR #116, opened 2019-06-23
add setup-otp for setting up OTP credentials for sudo

cg505:
This automates the procedure for adding OTP credentials to the puppet private share. It requires ocf/puppet#682.

This script is intended to be used on a desktop, with a YubiKey plugged in. It will configure the YubiKey and upload the auth to Yubico's servers. Then, it will take the public ID that identifies the YubiKey and send it over to lightning, which will add it to the puppet OTP private share as well as set up the google authenticator secrets.

We incrementally use the public IDs for the YubiKeys that start with `vvfuckcrv` for easy identification of OCF OTPs. There is no potential for public ID collisions because the Yubico servers do not allow a public ID to be replaced by another Yubikey with the same public ID.

### DIFF
--- a/acct/setup-otp
+++ b/acct/setup-otp
@@ -1,0 +1,59 @@
+#!/usr/bin/env bash
+
+set -euo pipefail
+
+if [ "$#" -ne 1 ]; then
+    echo "Usage: setup-otp <username>"
+    exit 1
+fi
+
+username="$1"
+
+# get last ocf-tagged public ID
+IDHex=$(sudo cut -d: -f2- /etc/otp/yubikeys | tr : \\n | grep vvfuckcrv | tr "cbdefghijklnrtuv" "0123456789ABCDEF" | sort | tail -1)
+
+while true; do
+    # increment to get next ID
+    IDHex=$(echo "obase=ibase=16;${IDHex}+1" | bc)
+    ID=$(echo "$IDHex" | tr "0123456789ABCDEF" "cbdefghijklnrtuv")
+    if curl -H "Content-Type: application/json" -d "{\"public_id\":\"$ID\",\"private_id\":\"000000000000\",\"aes_key\":\"00000000000000000000000000000000\",\"serial\":0}" 'https://upload.yubico.com/prepare' 2> /dev/null | grep -v "PUBLIC_ID_OCCUPIED" >/dev/null; then
+        # this public ID is not taken
+        break
+    fi
+done
+
+# yubikey generate script
+echo "Make sure your Yubikey is connected."
+echo "Programming with public ID $ID."
+read -rp "Is this a FRESH Yubikey (OTP slots unconfigured)? [yes/no] " fresh
+ykman config usb -e OTP
+case $fresh in
+yes)
+    echo "[The factory OTP will be left in slot 2.]"
+    ykman otp swap
+    slot=1
+    ;;
+no)
+    read -rp "Which OTP slot should I use? (1/2) " slot
+    ;;
+*)
+    echo "i said 'yes' or 'no', none of this $fresh nonsense smh"
+    exit 1
+    ;;
+esac
+
+case $slot in
+[12])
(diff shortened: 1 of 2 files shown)


acct/setup-otp
```diff
@@ -0,0 +1,58 @@
...
+    fi
+done
+
+# yubikey generate script
+echo "Make sure your Yubikey is connected."
+echo "Programming with public ID $ID."
+read -rp "Is this a FRESH Yubikey (OTP slots unconfigured)? [yes/no] " fresh
+case $fresh in
+yes)
+    echo "[The factory OTP will be left in slot 2.]"
```
cg505: Should I add a command to enable OTP here if the feature is disabled? (`ykman config usb -e OTP`)

dkess: Is it idempotent?

acct/setup-otp
```diff
@@ -0,0 +1,58 @@
...
+    ykman otp swap
+    slot=1
+    ;;
+no)
+    read -rp "Which OTP slot should I use? (1/2) " slot
+    ;;
+*)
+    echo "i said 'yes' or 'no', none of this $fresh nonsense smh"
```
jvperrin: Lol!

jvperrin:
Also maybe put "ocf" in that public ID somewhere instead for identification? Seems like a better message to send :)

acct/setup-otp
cg505: I think so.

cg505:
@jvperrin 
>How does an app like google authenticator or authy get configured?

Once you ssh to lightning, the script there will show a QR code/secret that you can use to configure your phone.

>Also maybe put "ocf" in that public ID somewhere instead for identification?

`crv` is `0cf` in modhex. Unfortunately `o` is not valid modhex so you can't put `ocf` in there directly. I basically just chose something memorable to put there to make it easily identifiable.

jvperrin:
I see, I didn't realize how modhex works, I guess this is memorable enough, idk

Anyway, that all sounds good to me

acct/setup-otp
```diff
@@ -0,0 +1,59 @@
...
+fi
+
+username="$1"
+
+# get last ocf-tagged public ID
+IDHex=$(sudo cut -d: -f2- /etc/otp/yubikeys | tr : \\n | grep vvfuckcrv | tr "cbdefghijklnrtuv" "0123456789ABCDEF" | sort | tail -1)
+
+while true; do
+    # increment to get next ID
+    IDHex=$(echo "obase=ibase=16;${IDHex}+1" | bc)
```
dkess: Here's to hoping we never run into a race condition with this, lol

cg505: The utility will fall hard if the public ID is taken, so if that happens you just rerun the script. The only thing is that you'll have to use slot 1 instead of "fresh".
However, a race on this line is not a problem because if I key is registered but not yet present in the file it will get skipped over by the curl incantation, and we'll increment again.

cg505:
holding on merging this until I can upload a newer version of the `yubikey-manager` package to our APT repo

cg505:
Uploaded a new version of `yubikey-manager` and dependencies that should have the features we need.

acct/setup-otp
```diff
@@ -0,0 +1,59 @@
...
+
+if [ "$#" -ne 1 ]; then
+    echo "Usage: setup-otp <username>"
+    exit 1
+fi
+
+username="$1"
+
+# get last ocf-tagged public ID
+IDHex=$(sudo cut -d: -f2- /etc/otp/yubikeys | tr : \\n | grep vvfuckcrv | tr "cbdefghijklnrtuv" "0123456789ABCDEF" | sort | tail -1)
```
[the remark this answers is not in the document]
kpengboy: Nit: in this repository, it seems the established convention is to use lowercase_with_underscores

cg505:
going to reopen this somewhat revamped after break

PR #117, opened 2019-06-23
ldap-lint: allow staffvm hozers

dkess:
Sometimes we give hozers an "owner" like a staffvm, but this causes ldap-lint to complain since they're not in the staffvm range. We should ignore those errors since it's still fine to do this.


dkess:
Right now we're getting daily rootspam for `hozer-87`, don't remember which staffer we gave it to but this was the easiest way to give them root for it.

jvperrin:
Ok, fair enough

PR #118, opened 2019-06-25
Prevent note script from running as root

MrMinos:


cg505:
Some people might assume it requires root and run it with sudo, which will not correctly mark which user ran note in User_Info.

However I do think it would be better to issue a warning and confirmation instead of failing hard. It makes sense that we could find ourselves in situations where we might want to do this.

MrMinos:
> getpass is already imported, you can use `getpass.getuser() == "root"`

nice!

> more importantly, this script should be safe to run as root, as I understand it. why do we need to block it?

What cg505 mentioned is the reason. I mistakenly ran it as root and messed up the User_Info file in conjunction with another issue.

> However I do think it would be better to issue a warning and confirmation instead of failing hard. It makes sense that we could find ourselves in situations where we might want to do this.

I would like to keep it simple. If the circumstances require `note` to be run as root, which I cannot think of any, we can modify the script on the fly. Certainly, such modification should be feasible given the necessity of root privilege to run `sudo note`.

staff/acct/note
```diff
@@ -34,6 +34,10 @@ def guess_topic(note):
 
 
 def main():
+    if getpass.getuser() == 'root':
+        print('Do not run this script as root. Aborting.')
+        return 6
```
cg505: `6`?

MrMinos: 3,4,5 was taken already in this script. no significant reason behind it.

abizer:
we should never have to modify the script on the fly to perform behavior we might expect. Why did User_Info get messed up when you ran it as root?

MrMinos:
The behavior we expect is never to run this as root. So there shouldn't be a situation where we need to run this as root.

abizer:
that's not correct - we have no expectations at all about how the script might be called. it's entirely reasonable that the script might be called by a user  from a root shell. the question still stands: why did it fail when you ran it as root? please figure that out.

ethanhs:
@abizer As I understand it, cooperc has already explained the issue, it loses history of who ran note. Seeing as root can do whatever they please, this change is tantamount to a warning. Perhaps an explicit warning and confirmation would be a good compromise? I agree that one should not need to edit the script to have it work how you want. That being said, I also see an argument that if root should not be running note, that shouldn't be a supported use case.

dkess:
I guess we could use the `$USER` environment variable instead, which the staffer could edit if they really wanted to.

ethanhs:
Or perhaps just prompt for the username if run as root and run as that user?

cg505:
The only situations in which you might want to run this as root are when OCF infra is heavily compromised; at this point it is totally reasonable to have to edit scripts to get them to work.

PR #123, opened 2019-07-30
gapps: add easyfilters script

dkess:
Feel free to comment/nitpick on anything in this PR, including how this is presented to the user.

### DIFF
--- a/staff/gapps/easyfilters
+++ b/staff/gapps/easyfilters
@@ -1,0 +1,6 @@
+#!/bin/bash -eu
+if [[ "$(hostname)" != "supernova" && "$(hostname)" != "dev-supernova" ]]; then
+    echo -e '\033[1;31mYou must run this command on supernova.\033[0m'
+    exit 1
+fi
+sudo -u ocfeasyfilters /opt/share/utils/staff/gapps/easyfilters-real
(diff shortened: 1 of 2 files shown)


jvperrin:
(I know we've talked about this several times now, but I still don't get why a script is needed in this case or what it really adds and it feels like you've jumped towards a solution without considering why we need it or what alternatives exist)

abizer:
to be fair, we've hashed out the UX of this many times over. this workflow is meant for when we create new user's GApps accounts for the first time, we can easily set them up with a set of default filters at creation time, rather than asking them to do it themselves using the filters XML, which requires several button clicks and navigation of the Gmail settings UI to upload. Here, we run easyfilters once we add their account information to LDAP, and they're set up. no more work on their part needed.

cg505:
We talked about this a bit last night: I think this is an excellent start but it would be even better if we can put as much as possible in this script: creating the user, adding to groups, adding filters, at least.

dkess:
To be clear: I don't believe it's possible for root staffers to run `easyfilters` on the user's behalf. The user who wants filters would have to run this on their own and authenticate with their account. So this script would be separate from any tools for adding people to Officers.

ethanhs:
I don't really recall where we ended up on this. I think I agree with @jvperrin on this, if we are expecting staff to run this script we may as well 

a) link to the filter XML
b) document how to add it/link to the filter page
c) maybe provide a way to customize it?

dkess:
This script was a bad idea. There are a few open-source tools that do general version of this, any of which would be better suited to our purpose. I use https://github.com/mbrt/gmailctl at work.

PR #124, opened 2019-08-28
makeservices: easywp delete tables option

keur:
Helpful if the user has an old instance of wordpress that they aren't
interested in.


keur:
fuck lint

PR #126, opened 2019-09-12
Update gen-rootpw to only print hash and use python

jvperrin:
This goes along with https://github.com/ocf/puppet/pull/751, since this script would otherwise be broken by that change since the root password hash is being moved to hieradata instead of a separate file in the private share.

I also decided to rewrite this in python since it was sufficiently complex and I wanted to do the hash confirmation stuff in python anyway as a sanity check.

Here's some examples of it being used, first with a password and confirmation matching (`test password` is the password used):
```
(venv) jvperrin@fireball:~/utils ‹master*›$ ./staff/puppet/gen-rootpw
################################################################################
This will generate a new password hash, but not update it across machines

To do that, you will need to put this new password hash into the puppet
private hiera file. Do NOT do this without input from others!
################################################################################

New password:
New password confirmation:
Here is the new password hash, go update ocf::root_password in puppet:
$6$rounds=65536$0FAqyBXIlxZ2ReR$NI/bFmjGGewVRC.lZc6hz.VlgLmLRXQiTO2wncUZOsO45cN.uqEbQF507DQ0k3z9mbC2rMRRTRIauxocvKwuM1
```

When the password and the confirmation do not match:
```
(venv) jvperrin@fireball:~/utils ‹update-gen-rootpw-script›$ ./staff/puppet/gen-rootpw
################################################################################
This will generate a new password hash, but not update it across machines

To do that, you will need to put this new password hash into the puppet
private hiera file. Do NOT do this without input from others!
################################################################################

New password:
New password confirmation:
Password and confirmation do not match, please try again
```


staff/puppet/gen-rootpw
```diff
@@ -1,28 +1,52 @@
...
+        print("by mkpasswd, make sure that there aren't any newlines, etc. in")
+        print('your password')
+        sys.exit(2)
+
+    print('Here is the new password hash, go update ocf::root_password in puppet:')
+    print(hashed_password)
+
+
+if __name__ == '__main__':
+    sys.exit(main())
```
ethanhs: Very minor nit but you `sys.exit` in `main()` as well as here. It might be nice to choose one, i.e. just `sys.exit(0)` at the end.

gnowxilef:
i'm curious, if this only ever outputs the hash, where is the actual root password located? or is the root pw never actually used?

ethanhs:
@gnowxilef In the hearts and minds of all root staffers :P

jvperrin:
@gnowxilef So the root password is remembered by the person who enters it into this script, who then passes it on to a small group of other staffers (SMs/DSMs generally), because the root password is very powerful and almost always not needed to solve problems (some exceptions being when LDAP/Kerberos are down and credential caches are empty for some reason, disk space is completely full on a box, or something is wrong with the sudoers file). Essentially the root password isn't supposed to be written down, especially electronically, unless it's encrypted or otherwise secured, but the hash is reasonably ok to put somewhere, although we'd still not want to make it public in case it can be cracked in some way.

TL;DR: We're reasonably paranoid about this password in particular, so only the hashed version is put anywhere apart from in people's brains and this hashed version put in the same place as other secrets to keep it secret too.

staff/puppet/gen-rootpw
jvperrin: Ah nice, yeah it should just be a `return` in the main function instead, thanks for noticing that

staff/puppet/gen-rootpw
```diff
@@ -1,28 +1,52 @@
...
+
+    password = getpass('New password: ')
+    password_confirm = getpass('New password confirmation: ')
+
+    if password != password_confirm:
+        print('Password and confirmation do not match, please try again')
+        sys.exit(1)
+
+    hashed_password = subprocess.check_output(
+        ('mkpasswd', '-m', 'sha-512', '-R', '65536', password),
```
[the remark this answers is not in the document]
jvperrin: Good idea, I did briefly think about this but forgot about it when making this version. I did dismiss it at the time since it would be very short-lived, but I definitely agree stdin would be much better for this, so I'll change that.

Thanks!

gnowxilef:
ahh, makes sense. thanks for the edification.

PR #127, opened 2019-09-17
Update signat to make it doesn't look bad

axmmisaka:
Initially it will look ugly (and makes `check` ugly because `check`
calls this - fortunately it's not function call otherwise it's gonna be more workload). This update makes it stop display ugly error on
non-supernova (or other machine in the future) machines.

It works as usual on supernova, but I'm considering if using uuid (/etc/machine-id) instead of hostname would be a better option.

Also, in the future `check` maybe rewritten in python and call functions in `signat` directly...? So do you think putting these stuff in program entrance would be a better idea...?

### DIFF
--- a/staff/acct/signat
+++ b/staff/acct/signat
@@ -5,6 +5,7 @@ student group.
 import argparse
 import sys
 import textwrap
+import platform
 
 import ocflib.account.search as search
 import ocflib.ucb.directory as directory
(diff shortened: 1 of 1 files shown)


dkess:
I'd make this a blacklist instead of a whitelist-- I'm able to get `signat` working on most staff-only hosts. I assume it only fails on tsunami?

Even better, simply make the script fail at the point where it actually needs to fail-- for example, if it fails because it can't find a particular credentials file, that's where the error message should show up.

axmmisaka:
It seems for me that it's broken on every computer except `supernova`. Can you, if possible, tell me how to do that?

dkess:
What command are you running? When I run `signat group open computing facility` I get normal-looking output.

axmmisaka:
It turns out that all machines except supernova are running newer ocflib, and the problem is parsing/whatever in ocflib is not compatible with older version.
Process from the very beginning through HTTP requesting orapps.berkeley.edu is fine, but parsing (group.py in ocflib) is where error occurs.

axmmisaka:
This is totally more complicated than what I thought it would be
https://github.com/ocf/utils/issues/128
Closing, will open new pr

PR #129, opened 2019-10-04
makeservices: makemysql automatic wp db pass

gundralaa:
Modified makemysql and makemysql-real to automatically update wp db pass if pass is reset

### DIFF
--- a/makeservices/makemysql
+++ b/makeservices/makemysql
@@ -5,5 +5,6 @@ if [[ "$(hostname)" != "tsunami" && "$(hostname)" != "dev-tsunami" ]]; then
 fi
 
 PASS=$(sudo -u mysql /opt/share/utils/makeservices/makemysql-real | tee /dev/tty | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
+
 cd ~/public_html/
-wp config set DB_PASSWORD "$PASS" > /dev/null 2>&1
+wp config set DB_PASSWORD "$PASS" > /dev/null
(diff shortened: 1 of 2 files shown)


dkess:
This approach breaks an important constraint of our current system-- we don't want users to be able to choose their database passwords.

How I would recommend doing it: `makemysql-real` generate the password, but capture the password from stdout (see https://github.com/ocf/utils/blob/master/makeservices/easywp#L60 for an example of how to do this), and pass that into `wp config`. This also means you won't have to juggle temporary files.

makeservices/makemysql
```diff
@@ -3,4 +3,7 @@ if [[ "$(hostname)" != "tsunami" && "$(hostname)" != "dev-tsunami" ]]; then
     echo -e '\033[1;31mYou must run this command on tsunami.\033[0m'
     exit 1
 fi
-sudo -u mysql /opt/share/utils/makeservices/makemysql-real
+
+PASS=$(sudo -u mysql /opt/share/utils/makeservices/makemysql-real | tee /dev/tty | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
+cd ~/public_html/
+wp config set DB_PASSWORD "$PASS" > /dev/null 2>&1
```
dkess: Why are you throwing out this error? Silent failures can create a lot of confusion.

gundralaa: I wanted to make sure the script didn't interfere with other scripts that use the makemysql command. Ill modify to only throw out the stdout

dkess: Hmm, that's a good point, stripping stdout is probably OK here.

abizer:
please make sure to squash commits before merging them.

also, makemysql is used in more contexts than /just/ updating WordPress passwords - I don't use WordPress for example, but I do use my OCF MySQL instance for other reasons. This is nominally safe, but should probably prompt to accept before running commands against someone's WordPress install.

Issue #121, opened 2019-07-13
Rewrite sorry/unsorry in python


jvperrin:
Currently [sorry](https://github.com/ocf/utils/blob/master/staff/acct/sorry/sorry) and [unsorry](https://github.com/ocf/utils/blob/master/staff/acct/sorry/unsorry) are pretty complicated scripts to be written in bash and have commented out sections, shellcheck-ignored warnings, etc. It's also tough to know if things done in the `sorry` script are undone properly in the `unsorry` script (setting file permissions for instance), so rewriting these two to use common functions and better error handling (maybe even be in the same script file) would be useful.

gnowxilef:
oh wow, unsorry/sorry is still that terrible bash script? nice.

jvperrin:
Yup, exactly, every time I have to look at it or use it and it breaks halfway through I get sad :'(

PR #130, opened 2019-10-05
Finished sorry in Python without testing

axmmisaka:
I don't have sudo, so I don't know if calling ldapmodify like this/shutil.chown/chmod/etc... will work, but I have tested (almost) everything I can test for sure.
https://github.com/ocf/utils/issues/121
This issue is meant to make bash script more clear and manageable and maintainable, however it seems that the script becomes more unreadable and uglier.

### DIFF
--- a/staff/acct/sorry/sorry
+++ b/staff/acct/sorry/sorry
@@ -1,156 +1,183 @@
-#!/bin/bash
-# Script for sorrying OCF user accounts
+#!/usr/bin/env python3
+# Script for sorrying OCF user accounts rewritten in python3
+import argparse
+import datetime
+import grp
+import os
+import shutil
+import subprocess
+import sys
 
-REASONPATH=/opt/share/utils/staff/acct/sorry
-LDAPSEARCH=$(which ldapsearch)
-LDAPMODIFY=$(which ldapmodify)
-KINIT=$(which kinit)
-KDESTROY=$(which kdestroy)
-#KRB5CCNAME=/root/krb5cc_sorry
-#export KRB5CCNAME
+import ocflib.account.manage
+import ocflib.account.search
+import ocflib.account.utils
+import ocflib.misc.mail
 
-# check to see if running as root
-if [ "$(/usr/bin/id -u)" != 0 ]; then
-    echo "You must be root to run this."
-    exit 2
-fi
+reason_path = '/opt/share/utils/staff/acct/sorry/'
+sorry_shell = '/opt/share/utils/bin/sorried'
 
-if [ -z "$1" ] || [ -z "$2" ]; then
-    echo "Usage: $0 [user to be sorried] [sorry reason file]"
-    echo
-    echo "Standard Sorry Reasons:"
-    find "$REASONPATH" -maxdepth 1 -mindepth 1 -type f -printf "    %f\n"
-    echo
-    echo "For custom sorry reasons you can pass in your own file"
-    exit 0
-fi
 
-sorriedUser=$1
+def find_sorry_file(file_path):
+    # Own sorry file
+    if (os.path.isfile(file_path)):
+        return file_path
 
-if [[ ! "$sorriedUser" =~ ^[a-z]+ ]]; then
-    echo "$sorriedUser is an invalid username"
-    exit 3
-fi
+    # Default sorry files
+    if (os.path.isfile(reason_path + file_path)):
+        return reason_path + file_path
 
-if [ -z "$(getent passwd "$sorriedUser")" ]; then
-    echo "User $sorriedUser does not exist"
-    exit 3
-fi
+    raise FileNotFoundError("FATAL: Sorry file \"{}\" does not exist. Exiting.".format(file_path))
 
-sorryFile=$2
 
-if [ ! -f "$sorryFile" ]; then
-    sorryFile="$REASONPATH/$sorryFile"
-    if [ ! -f "$sorryFile" ]; then
-        echo "Invalid sorry file"
-        exit 3
-    fi
-fi
+def sorry_user(username, sorry_file):
+    attributes = ocflib.account.search.user_attrs(username)
+    if attributes is None:
+        raise ValueError('User not found.')
 
-userdir=$(ldapsearch -x uid="$sorriedUser" | grep homeDirectory | cut -d' ' -f2)
+    rootstaffer = real_user()
+    kerberos_credentials(rootstaffer)
 
-# Rewriting this to use bash variable substitution is annoying and confusing, so
-# just use sed instead and ignore the shellcheck warning
-# shellcheck disable=SC2001
-httpdir=$(echo "$sorriedUser" | sed -E 's%([a-z])[a-z]*%/services/http/users/\1/\0%')
+    user_dir = attributes['homeDirectory']
 
-rootstaffer="$SUDO_USER"
+    # Workaround for ldap3 1.x.x machines
+    if isinstance(user_dir, list):
+        user_dir = user_dir[0]
 
-if [ "$rootstaffer" = "root" ] || [ -z "$rootstaffer" ]; then
-    echo "The sorry.log is much more useful when it logs who you are"
-    echo "rather than simply 'root'. Please enter your username:"
-    read -r rootstaffer
-fi
+    print('Copying sorry file and making .oldshell file')
 
-if [ -z "$SORRY_KRB5CCNAME" ]; then
-    echo "You are $rootstaffer"
-    $KINIT "${rootstaffer}/admin"
-    if [ $? -ne 0 ]; then
-        echo "kinit failed, bailing out!"
-        exit 1
-    fi
-else
-    echo "SORRY_KRB5CCNAME set in environment."
-    echo "Assuming this file contains current admin credentials."
-    KRB5CCNAME="$SORRY_KRB5CCNAME"
-    export KRB5CCNAME
-fi
+    if not (os.path.isfile(user_dir + '/.oldshell')):
+        oldshell = attributes['loginShell']
+        # Workaround for ldap3 1.x.x machines
+        if isinstance(oldshell, list):
+            oldshell = user_dir[0]
 
-sorryshell=/opt/share/utils/bin/sorried
+        oldshell_file = open(user_dir + '/.oldshell', 'w')
+        oldshell_file.write(oldshell)
+        oldshell_file.close()
 
-echo ""
-echo "Copying sorry file and making .oldshell file"
+    shutil.copy2(sorry_file, user_dir + '/.sorry')
+    os.chmod(user_dir + '/.sorry', 0o400)
+    shutil.chown(user_dir + '/.sorry', user=username, group='ocf')
 
-oldshell=$($LDAPSEARCH -x "(uid=$sorriedUser)" loginShell | grep "^loginShell:" | cut -d" " -f2)
+    print("Changing user's shell to sorry shell")
 
-if [[ ! -f .oldshell ]]; then
-    echo "$oldshell" > "$userdir/.oldshell"
-fi
+    ocflib.infra.ldap.modify_ldap_entry(ocflib.account.utils.dn_for_username(username),
+                                        {'loginShell': sorry_shell, 'gidNumber': grp.getgrnam('sorry').gr_gid})
 
-cp "$sorryFile" "$userdir/.sorry"
-chmod 400 "$userdir/.sorry"
-chown "$sorriedUser:ocf" "$userdir/.sorry"
+    http_dir = '/services/http/users/{0}/{1}'.format(username[0], username)
+    print('Changing permissions on httpdir to 000')
+    if (os.path.isdir(http_dir)):
+        os.chmod(http_dir, 0o000)
 
-echo ""
-echo "Changing user's shell to a sorry shell"
+    print('Changing permissions on home directory to 500')
+    os.chmod(user_dir, 0o500)
 
-sorrygid=$(getent group sorry | cut -d : -f 3)
-$LDAPMODIFY -H ldaps://ldap.ocf.berkeley.edu <<EOF
-dn: uid=$sorriedUser,ou=People,dc=OCF,dc=Berkeley,dc=EDU
-changetype: modify
-replace: loginShell
-loginShell: $sorryshell
--
-replace: gidNumber
-gidNumber: $sorrygid
-EOF
+    # WTF is final system check for?
+    # log(rootstaffer, username)
+    send_log_mail(rootstaffer, username, open(reason_path, 'r').read())
 
-echo "Changing permissions on httpdir to 000"
-if [[ -d "$httpdir" ]]; then
-    chmod 000 "$httpdir"
-fi
+    user_mail = get_user_real_mail(username)
+    if user_mail:
+        send_sorry_mail(user_mail, username, open(reason_path, 'r').read())
+    kerberos_destroy()
 
-echo ""
-echo "Changing permissions on home directory to 500"
-chmod 500 "$userdir"
+# Returns the real user who is running this script
 
-echo ""
-echo ""
-echo ""
-echo "Final system check"
-ldapsearch -x uid="$sorriedUser"
-ldapsearch -x cn=sorry | tail
-finger -m "$sorriedUser"
-ls -la "$userdir"
-ls -ld "$httpdir"
 
-# logging
-echo "$(/bin/date) - $rootstaffer $sorriedUser" >> /opt/acct/sorry.log
+def real_user():
+    if 'SUDO_USER' in os.environ and os.environ['SUDO_USER'] != 'root':
+        return os.environ['SUDO_USER']
+    else:
+        print("The sorry.log is much more useful when it logs who you are rather than simply 'root'.")
+        username = input('Please enter your username: ')
+        return username
 
-# Notify user by email if email address is available.
-email_from='Open Computing Facility <help@ocf.berkeley.edu>'
-email_subject='[OCF] Account disabled'
-email_to="$(ldapsearch "uid=${sorriedUser}" mail | grep ^mail | cut -d' ' -f2)"
-[ -n "$email_to" ] && \
-  mail -a "From: ${email_from}" -s "$email_subject" "$email_to" <<EOF
+
+def kerberos_credentials(staff_name):
+    if ('SORRY_KRB5CCNAME' in os.environ):
+        print('SORRY_KRB5CCNAME set in environment. Assuming this file contains current admin credentials.')
+        os.environ['KRB5CCNAME'] = os.environ['SORRY_KRB5CCNAME']
+    else:
+        print('Your are {0}.'.format(staff_name))
+        if (subprocess.call([shutil.which('kinit'), '{0}/admin'.format(staff_name)]) != 0):
+            raise OSError('Kinit failed, bailing out!')
+        else:
+            print('Kinit complete.')
+
+
+def kerberos_destroy():
+    return subprocess.call(['kdestroy'])
+
+
+def get_user_real_mail(username):
+    email = str(subprocess.run([shutil.which('ldapsearch'),
+                                'uid={0}'.format(username), 'mail'], stdout=subprocess.PIPE).stdout)
+    email = email[email.find('\\nmail') + 8:]
+    email = email[:email.find('\\n\\n')]
+    return email
+
+
+def log(root_staffer, sorried_user):
+    with open('/opt/acct/sorry.log', 'a') as log_file:
+        # Date style matters...?
+        log_file.write('{0} - {1} {2}'.format(str(datetime.datetime.now()), root_staffer, sorried_user))
+
+
+def send_log_mail(root_staffer, sorried_user, reason):
+    email_from = 'Open Computing Facility <sorry_log@ocf.berkeley.edu>'
+    email_subject = '[OCF] Account disabled'
+    email_body = '''
+This is an automated email sent by sorry script. This is the log of a recent sorry operation:
+Account "{0}" has been disabled by "{1}" for reason "{2}" on "{3}".
+'''.format(sorried_user, root_staffer, reason, str(datetime.datetime.now()))
+    ocflib.misc.mail.send_mail('root@ocf.berkeley.edu', email_subject, email_body, cc=None, sender=email_from)
+
+
+def send_sorry_mail(sorried_user_email, username, reason):
+    email_from = 'Open Computing Facility <help@ocf.berkeley.edu>'
+    email_subject = '[OCF] Account disabled'
+    email_body = '''
 Hello,
 
 Your OCF account has been disabled.
 
-Your account name is: ${sorriedUser}
+Your account name is: {0}
 
 The reason your account has been disabled:
-$(cat "$sorryFile")
+{1}
 
 Feel free to reply to this message.
 Please do not share your password with anyone or over email.
 
-$(python3 -c 'import ocflib.misc.mail as mail; print(mail.MAIL_SIGNATURE)')
-EOF
+{2}
+    '''.format(username, reason, ocflib.misc.mail.MAIL_SIGNATURE)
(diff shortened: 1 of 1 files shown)


staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+        log_file.write('{0} - {1} {2}'.format(str(datetime.datetime.now()), root_staffer, sorried_user))
+
+
+def send_log_mail(root_staffer, sorried_user, reason):
+    email_from = 'Open Computing Facility <sorry_log@ocf.berkeley.edu>'
+    email_subject = '[OCF] Account disabled'
+    email_body = '''
+This is an automated email sent by sorry script. This is the log of a recent sorry operation:
+Account "{0}" has been disabled by "{1}" for reason "{2}" on "{3}".
+'''.format(sorried_user, root_staffer, reason, str(datetime.datetime.now()))
```
jvperrin: `dedent` (`from textwrap import dedent`) would be useful here so you can intent all this more, I also mention it further down. I do like the idea of logging to email like this, although I'm not sure how noisy it'd be. I doubt we sorry accounts often enough for it to be too annoying though.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+            raise OSError('Kinit failed, bailing out!')
+        else:
+            print('Kinit complete.')
+
+
+def kerberos_destroy():
+    return subprocess.call(['kdestroy'])
+
+
+def get_user_real_mail(username):
```
jvperrin: I've written a function for this in [update-email](https://github.com/ocf/utils/blob/6e7c2ac6daa39d62c875af3378896d5ffcd606f9/acct/update-email#L52-L67), but maybe it should be put into ocflib so that both that and this script can use it? It does need [some kerberos-related stuff to be run first](https://github.com/ocf/utils/blob/6e7c2ac6daa39d62c875af3378896d5ffcd606f9/acct/update-email#L89-L100) though, similar to this script.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+
+    http_dir = '/services/http/users/{0}/{1}'.format(username[0], username)
+    print('Changing permissions on httpdir to 000')
+    if (os.path.isdir(http_dir)):
+        os.chmod(http_dir, 0o000)
+
+    print('Changing permissions on home directory to 500')
+    os.chmod(user_dir, 0o500)
+
+    # WTF is final system check for?
```
jvperrin: Honestly I think it's just to show some debugging information since this bash script is not the most stable and often has errors as it goes, so that's why that part is there and it can probably be removed on this python version.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+    ocflib.infra.ldap.modify_ldap_entry(ocflib.account.utils.dn_for_username(username),
+                                        {'loginShell': sorry_shell, 'gidNumber': grp.getgrnam('sorry').gr_gid})
+
+    http_dir = '/services/http/users/{0}/{1}'.format(username[0], username)
+    print('Changing permissions on httpdir to 000')
+    if (os.path.isdir(http_dir)):
+        os.chmod(http_dir, 0o000)
+
+    print('Changing permissions on home directory to 500')
+    os.chmod(user_dir, 0o500)
```
jvperrin: We have a couple ocflib functions that might help a little bit here:

- `home_dir` from `ocflib.account.utils` to get a user's home directory path
- `web_dir` from `ocflib.account.utils` for a user's web directory

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+    else:
+        print("The sorry.log is much more useful when it logs who you are rather than simply 'root'.")
+        username = input('Please enter your username: ')
+        return username
+
+
+def kerberos_credentials(staff_name):
+    if ('SORRY_KRB5CCNAME' in os.environ):
+        print('SORRY_KRB5CCNAME set in environment. Assuming this file contains current admin credentials.')
+        os.environ['KRB5CCNAME'] = os.environ['SORRY_KRB5CCNAME']
```
jvperrin: In practice I'm not sure this path is ever used, or at least I've never heard of it being used by anyone in recent history. Maybe it can be removed in favor of just running `kinit` each time if a kerberos ticket doesn't already exist?

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+    else:
+        print('Your are {0}.'.format(staff_name))
+        if (subprocess.call([shutil.which('kinit'), '{0}/admin'.format(staff_name)]) != 0):
```
cg505: Ideally, if we're already authenticated with an admin principal, we shouldn't call kinit. `klist --json` will give you a JSON dump of the current credential cache. If it contains the k/v pair `"principal": "username/admin@OCF.BERKELEY.EDU"`, then we don't need to do kinit. This could look something like this:

```python
import json
...
klist = subprocess.run(['sudo', '-u', staff_name, 'klist', '--json'], stdout=subprocess.PIPE)
if klist.returncode == 0:
    cache_info = json.loads(klist.stdout.decode())
    if cache_info.get('principal') == '{0}/admin@OCF.BERKELEY.EDU'.format(staff_name):
        print('Reading from credential cache.')
        return
... #continue to kinit
```

The `sudo` bit is necessary because this script is running as root but `klist` needs to run as the user.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+            raise OSError('Kinit failed, bailing out!')
+        else:
+            print('Kinit complete.')
+
+
+def kerberos_destroy():
+    return subprocess.call(['kdestroy'])
```
cg505: Ideally we don't do this if we didn't run kinit (e.g. if using `SORRY_KRB5CCNAME`).

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+
+    if os.getuid() != 0:
+        raise PermissionError('You must be root to run this.')
+
+    args = parser.parse_args()
+    file_path = find_sorry_file(args.sorryReason)
+    print('Sorry file {}.'.format(file_path))
+
+    sorry_user(args.sorriedUser, file_path)
+    print('NOW RUN note -u {0} reason. NOW'.format(args.sorriedUser))
```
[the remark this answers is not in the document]
cg505: It would be good to do this automatically. We should keep in mind that `note` does not have a way to read the reason from stdin by default, and the `reason` here should usually have more detail than the reason given in the sorryfile.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+    print('Changing permissions on httpdir to 000')
+    if (os.path.isdir(http_dir)):
+        os.chmod(http_dir, 0o000)
+
+    print('Changing permissions on home directory to 500')
+    os.chmod(user_dir, 0o500)
+
+    # WTF is final system check for?
+    # log(rootstaffer, username)
+    send_log_mail(rootstaffer, username, open(reason_path, 'r').read())
```
cg505:
```suggestion
    send_log_mail(rootstaffer, username, open(sorry_file, 'r').read())
```

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+
+    user_mail = get_user_real_mail(username)
+    if user_mail:
+        send_sorry_mail(user_mail, username, open(reason_path, 'r').read())
```
cg505:
```suggestion
        send_sorry_mail(user_mail, username, open(sorry_file, 'r').read())
```

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+        # Date style matters...?
+        log_file.write('{0} - {1} {2}'.format(str(datetime.datetime.now()), root_staffer, sorried_user))
+
+
+def send_log_mail(root_staffer, sorried_user, reason):
+    email_from = 'Open Computing Facility <sorry_log@ocf.berkeley.edu>'
+    email_subject = '[OCF] Account disabled'
+    email_body = '''
+This is an automated email sent by sorry script. This is the log of a recent sorry operation:
+Account "{0}" has been disabled by "{1}" for reason "{2}" on "{3}".
```
cg505: Usually the "reason" is multiple lines long, it should probably be on its own line.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+
+
+def log(root_staffer, sorried_user):
+    with open('/opt/acct/sorry.log', 'a') as log_file:
+        # Date style matters...?
+        log_file.write('{0} - {1} {2}'.format(str(datetime.datetime.now()), root_staffer, sorried_user))
+
+
+def send_log_mail(root_staffer, sorried_user, reason):
+    email_from = 'Open Computing Facility <sorry_log@ocf.berkeley.edu>'
```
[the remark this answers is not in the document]
cg505: well... it actually works right now, but that's not intended behavior and is not permanent.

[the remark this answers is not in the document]
axmmisaka: It’s technically just a header which you can even set to donald@wh.gov... The real sender is still root@o.b.e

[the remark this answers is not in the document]
cg505: iirc our mailserver will filter based on that header, and reject mail coming from addresses that don't exist

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+            print('Kinit complete.')
+
+
+def kerberos_destroy():
+    return subprocess.call(['kdestroy'])
+
+
+def get_user_real_mail(username):
+    email = str(subprocess.run([shutil.which('ldapsearch'),
+                                'uid={0}'.format(username), 'mail'], stdout=subprocess.PIPE).stdout)
```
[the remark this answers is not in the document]
axmmisaka: This runs fine, what will change if we use .decode()?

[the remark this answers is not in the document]
cg505: Casting bytes to string like this gives you this terrible thing that looks like
```
"b'# extended LDIF\\n#\\n# LDAPv3\\n# base <dc=ocf,dc=berkeley,dc=edu> (default) with scope subtree\\n# filter: uid=cooperc\\n# requesting: mail \\n#\\n\\n# cooperc, People, OCF.Berkeley.EDU\\ndn: uid=cooperc,ou=People,dc=OCF,dc=Berkeley,dc=EDU\\nmail: cooperc@g.ocf.berkeley.edu\\n\\n# search result\\nsearch: 5\\nresult: 0 Success\\n\\n# numResponses: 2\\n# numEntries: 1\\n'"
```
Using `.decode()`, which is the best practice way to convert bytes to str will give
```
'# extended LDIF\n#\n# LDAPv3\n# base <dc=ocf,dc=berkeley,dc=edu> (default) with scope subtree\n# filter: uid=cooperc\n# requesting: mail \n#\n\n# cooperc, People, OCF.Berkeley.EDU\ndn: uid=cooperc,ou=People,dc=OCF,dc=Berkeley,dc=EDU\nmail: cooperc@g.ocf.berkeley.edu\n\n# search result\nsearch: 5\nresult: 0 Success\n\n# numResponses: 2\n# numEntries: 1\n'
```

Basically `str()` will literally give you the string representation of the bytes, rather than actually converting it.

staff/acct/sorry/sorry
axmmisaka: It's in the original file which I have no clue what it does. May be an ocf legacy...

staff/acct/sorry/sorry
axmmisaka: Probably write kerberos_auth and update_email stuff in ocflib...?

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,183 @@
...
+import ocflib.account.utils
+import ocflib.misc.mail
+
+reason_path = '/opt/share/utils/staff/acct/sorry/'
+sorry_shell = '/opt/share/utils/bin/sorried'
+
+
+def find_sorry_file(file_path):
+    # Own sorry file
+    if (os.path.isfile(file_path)):
```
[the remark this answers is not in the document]
axmmisaka: Some C habits...

staff/acct/sorry/sorry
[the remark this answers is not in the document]
axmmisaka: I'll use jvperrin's function directly...

axmmisaka:
Sadly, I gotta do my own works and do the TODOs a moment later, probably....

staff/acct/sorry/sorry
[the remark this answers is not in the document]
axmmisaka: Maybe that's only for inbox?

staff/acct/sorry/sorry
```diff
@@ -74,12 +69,15 @@ def sorry_user(username, sorry_file):
...
-    send_log_mail(rootstaffer, username, open(reason_path, 'r').read())
+    with open(sorry_file, 'r') as filep:
+        send_log_mail(rootstaffer, username, filep.read())
 
     user_mail = get_user_real_mail(username)
     if user_mail:
-        send_sorry_mail(user_mail, username, open(reason_path, 'r').read())
-    kerberos_destroy()
+        with open(sorry_file, 'r') as filep:
+            send_sorry_mail(user_mail, username, filep.read())
```
cg505: we may as well just put this whole block in the same `with` statement from directly above

staff/acct/sorry/sorry
```diff
@@ -93,48 +91,62 @@ def real_user():
         return username
 
 
-def kerberos_credentials(staff_name):
+def kerberos_init(staff_name):
     if ('SORRY_KRB5CCNAME' in os.environ):
         print('SORRY_KRB5CCNAME set in environment. Assuming this file contains current admin credentials.')
         os.environ['KRB5CCNAME'] = os.environ['SORRY_KRB5CCNAME']
+        # TODO: What happened?
```
cg505, line 98: I don't think you need to do anything else here, this is fine.

axmmisaka:
I'll mess around with ocflib soon

staff/acct/sorry/sorry
```diff
@@ -90,12 +95,15 @@ def real_user():
         username = input('Please enter your username: ')
         return username
 
+# Initiates a kerberos ticket for current user
+# Returns True when a new ticket is created via kinit so that we destroy it afterwards
+# Raises OSError is kinit screwed up
```
cg505, line 100:
```suggestion
# Raises OSError if kinit screwed up
```

staff/acct/sorry/sorry
```diff
@@ -15,8 +16,10 @@ import ocflib.account.search
 import ocflib.account.utils
 import ocflib.misc.mail
 
-REASON_PATH = '/opt/share/utils/staff/acct/sorry/'
-SORRY_SHELL = '/opt/share/utils/bin/sorried'
+REASON_PATH = pathlib.Path('/opt/share/utils/staff/acct/sorry/')
+SORRY_SHELL = pathlib.Path('/opt/share/utils/bin/sorried')
+# TODO: This sorry log file does not exist anymore, do we still log to file?
+SORRY_LOG = pathlib.Path('/opt/acct/sorry.log')
```
cg505, line 22: i think we should probably reinstate this file, it would be useful

cg505: probably put it in NFS though

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+    print('Changing permissions on httpdir to 000')
+    if (os.path.isdir(http_dir)):
+        os.chmod(http_dir, 0o000)
+
+    print('Changing permissions on home directory to 500')
+    os.chmod(user_dir, 0o500)
+
+    # Omitting final system check for now
+    with open(sorry_file, 'r') as filep:
+        log(rootstaffer, username, filep.read())
```
kpengboy, line 75: `filep.read()` can only be called once; any subsequent calls would return `''`. You probably want to save its result to a variable and use that.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+        user_mail = ocflib.account.utils.get_email(username)
+        if user_mail:
+            send_sorry_mail(user_mail, username, filep.read())
+        choice = input("Do you want this script to run 'note -u {0} reason' for you? [Y/n]".format(username))
+        if choice not in ['n', 'N']:
+            subprocess.run(['sudo', '-u', rootstaffer, 'note', '-u', format(username), filep.read()])
```
kpengboy, line 81: This might be problematic if the sorry reason is more than one line. (And even for one-line reasons, you probably want to strip off any end-of-line character that you read.)

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+        kerberos_destroy()
+
+
+# Returns the real user who is running this script, but probably only works on Linux
+def real_user():
+    if 'SUDO_USER' in os.environ and os.environ['SUDO_USER'] != 'root':
+        return os.environ['SUDO_USER']
+    else:
+        print("The sorry.log is much more useful when it logs who you are rather than simply 'root'.")
+        username = input('Please enter your username: ')
```
kpengboy, line 95: Perhaps just ask "Please enter your username (for logging purposes): "?

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+        return username
+
+# Initiates a kerberos ticket for current user
+# Returns True when a new ticket is created via kinit so that we destroy it afterwards
+# Raises OSError is kinit screwed up
+
+
+def kerberos_init(staff_name):
+    if ('SORRY_KRB5CCNAME' in os.environ):
```
kpengboy, line 104: Nit: remove parentheses

staff/acct/sorry/sorry
kpengboy: I can't think of anything that uses this either; it's probably safe to remove. If we perchance discover that something actually uses it we can add it back...

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+    else:
+        print('You are {0}.'.format(staff_name))
+        klist = subprocess.run(['sudo', '-u', staff_name, 'klist', '--json'], stdout=subprocess.PIPE)
+        if klist.returncode == 0:
+            cache_info = json.loads(klist.stdout.decode())
+            if cache_info.get('principal') == '{0}/admin@OCF.BERKELEY.EDU'.format(staff_name):
+                print('Using your preexisting credentials.')
+                return False
+
+        if (subprocess.call(['kinit', '{0}/admin'.format(staff_name)]) != 0):
```
kpengboy, line 117: idem

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+    with open(SORRY_LOG, 'a') as log_file:
+        # Date style matters...?
+        log_file.write('{0} - {1} {2}'.format(str(datetime.datetime.now()), root_staffer, sorried_user))
+
+
+def send_log_mail(root_staffer, sorried_user, reason):
+    # Use root@ for now
+    email_from = 'Open Computing Facility <root@ocf.berkeley.edu>'
+    email_subject = '[OCF] Account disabled'
+    email_body = textwrap.dedent('''
```
kpengboy, line 149: Nit here and throughout: per PEP8, triple quotes should always be `"""`, not `'''`.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+    sorry_description = textwrap.dedent('''
+        Sorry a user.
+        Standard sorry reasons are:
+        {0}
+        Nevertheless, you can also pass in your own file with its file name as argument sorry_reason.
+    '''.format('\n'.join([f for f in os.listdir(REASON_PATH) if os.path.isfile(os.path.join(REASON_PATH, f))])))
+    parser = argparse.ArgumentParser(usage='%(prog)s sorried_user sorry_reason',
+                                     formatter_class=argparse.RawDescriptionHelpFormatter,
+                                     description=sorry_description)
+    parser.add_argument('sorried_user', metavar='sorried_user', help='Username of sorried user')
```
kpengboy, line 187: Nit: would put the metavars in ALL_CAPS.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+    parser.add_argument('sorry_reason', metavar='sorried_reason', help='Reason for the user to be sorried')
+
+    if os.getuid() != 0:
+        raise PermissionError('You must be root to run this.')
+
+    args = parser.parse_args()
+    file_path = find_sorry_file(args.sorryReason)
```
kpengboy, line 194:
```
sorryReason
```

Shouldn't this be `sorry_reason`?

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+    print('Sorry file {}.'.format(file_path))
+
+    # TODO: Help user to note... by sub-processing note -u user reason, maybe?
+    sorry_user(args.sorriedUser, file_path)
```
kpengboy, line 198:
```
sorriedUser
```

idem

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+        This is an automated email sent by OCF sorry script. This is the log of a recent sorry operation:
+        On "{3}", account "{0}" has been disabled by "{1}" for reason:
+        "{2}".
+        '''.format(sorried_user, root_staffer, reason, str(datetime.datetime.now())))
+    ocflib.misc.mail.send_mail('root@ocf.berkeley.edu', email_subject, email_body, cc=None, sender=email_from)
+
+
+def send_sorry_mail(sorried_user_email, username, reason):
+    email_subject = '[OCF] Account disabled'
+    email_body = '''
```
kpengboy, line 159: Nit: optionally consider using `textwrap.dedent`.

staff/acct/sorry/sorry
```diff
@@ -1,156 +1,202 @@
...
+The reason your account has been disabled is:
+{1}
 
 Feel free to reply to this message.
 Please do not share your password with anyone or over email.
 
-$(python3 -c 'import ocflib.misc.mail as mail; print(mail.MAIL_SIGNATURE)')
-EOF
+{2}
+    '''.format(username, reason, ocflib.misc.mail.MAIL_SIGNATURE)
```
kpengboy, line 173:
```suggestion
{2}'''.format(username, reason, ocflib.misc.mail.MAIL_SIGNATURE)
```

Avoids empty space at the end of the message.

staff/acct/sorry/sorry
axmmisaka: Will change this, I personally don't think that is what things are supposed to be - this might actually should be the filename of sorry reason

staff/acct/sorry/sorry
axmmisaka: That's copied from the original script, but will do

staff/acct/sorry/sorry
axmmisaka: Let me double check, i moved some code from `main()` to there

staff/acct/sorry/sorry
axmmisaka: This is probably a legacy before ocf transit all its code to github

staff/acct/sorry/sorry
cg505: Yeah, I think using the filename makes more sense.

ethanhs:
Would you be interested in resurrecting this PR? If not I can take a look at it.

axmmisaka:
> 
> 
> Would you be interested in resurrecting this PR? If not I can take a look at it.

Absolutely!

axmmisaka:
Closing https://github.com/ocf/utils/pull/130 https://github.com/ocf/utils/pull/152 https://github.com/ocf/ocflib/pull/210 https://github.com/ocf/utils/pull/153 altogether as I created a looping "merge dependency" almost impossible to review and merge/do CICD; I'll rewrite everything again this semester

PR #136, opened 2020-02-09
Add a ceph backing option to makevm

fydai:


sbin/add-to-group
```diff
@@ -0,0 +1 @@
+../staff/acct/add-to-group
```
cg505: should be a separate PR but I don't really care enough

staff/sys/makevm
```diff
@@ -262,7 +276,12 @@ def _main(args):
     print('\tOS Type: {}'.format(args.os_type))
     print('\tOS Variant: {}'.format(args.os_variant))
     print('\tDisk Space: {} GB'.format(args.storage))
-    print('\tVolume Group: {}'.format(args.vg))
+    if args.ceph:
```
cg505: we should throw an exception if we are passed both args.

staff/sys/makevm
```diff
@@ -201,9 +212,13 @@ def _main(args):
...
     parser.add_argument('-V', '--vg', type=str, default='vg',
                         help='LVM volume group to use for storage')
+    parser.add_argument('-C', '--ceph', action='store_true', default=False,
+                        help='use Ceph storage instead of LVM')
+    parser.add_argument('-P', '--ceph-pool', type=str, default='vm',
+                        help='Ceph Pool to use for storage')
     parser.add_argument('--os-type', type=str, default='linux',
                         help='os type')
-    parser.add_argument('--os-variant', type=str, default='debian9')
+    parser.add_argument('--os-variant', type=str, default='debian10')
```
cg505: I thought that debian10 didn't exist in virt-install yet?

staff/sys/makevm
```diff
@@ -242,6 +257,27 @@ def _main(args):
...
             sys.exit(2)
 
+    if args.ceph and not shutil.which('ceph'):
+        print("Warning: You are trying to use Ceph but I couldn't find the `ceph` executable.")
+
+        if not confirm():
+            print('Cancelled.')
+            sys.exit(2)
+
+    if args.ceph and args.vg != 'vg':
```
cg505: we should store the default value in a global variable or something so we only have to change it in one place

staff/sys/makevm
```diff
@@ -242,6 +257,27 @@ def _main(args):
...
+        print('Warning: You have Ceph backing storage, but specified a non-default VG.')
+
+        if not confirm():
+            print('Cancelled.')
+            sys.exit(2)
+
+    if not args.ceph and args.ceph_pool != 'vm':
```
cg505: same as above

staff/sys/makevm
ja5087: use `parser.get_default`

staff/sys/makevm
ja5087: `osinfo-query os | grep Debian` seems to indicate yes on pandemic but not riptide so maybe it's a buster thing?

cg505: we probably shouldn't change the default yet in that case.

PR #138, opened 2020-02-21
[OCF/util #137] easywp broken 

### LINKED ISSUE #137: easywp broken
After #129 was merged, easywp does not work on new accounts because makemysql fails (since there is no existing config to update the DB password in).

expedited:
Fixes #137

Problem:
We'd produce an error after trying to update password if WordPress is not already installed

Changes: 
Added check to see if wordpress is installed before running the password update command.

### DIFF
--- a/makeservices/makemysql
+++ b/makeservices/makemysql
@@ -8,4 +8,9 @@ PASS=$(sudo -u mysql /opt/share/utils/makeservices/makemysql-real | tee /dev/tty
 
 echo 'Changing WordPress database password'
 cd ~/public_html/
-wp config set DB_PASSWORD "$PASS" > /dev/null
+FILE=wp-includes/version.php
+if [ -f "$FILE" ]; then
+    wp config set DB_PASSWORD "$PASS" > /dev/null
+else
+    echo "WordPress is not installed!"
+fi


kpengboy:
I investigated; the `easywp` script is expecting the last line of `makemysql` to contain the password, but this assumption is no longer true.

We should probably fix that script to look for a line with the text "Your MySQL database password is:", or even better, add a mode to makemysql which only prints out the password (maybe printing other info to stderr).

makeservices/makemysql
```diff
@@ -8,4 +8,9 @@ PASS=$(sudo -u mysql /opt/share/utils/makeservices/makemysql-real | tee /dev/tty
 
 echo 'Changing WordPress database password'
 cd ~/public_html/
-wp config set DB_PASSWORD "$PASS" > /dev/null
+FILE=wp-includes/version.php
+if [ -f "$FILE" ]; then
+    wp config set DB_PASSWORD "$PASS" > /dev/null
+else
+    echo "WordPress is not installed!"
```
kpengboy, line 15:
```suggestion
    echo "WordPress is not installed."
```
Nit: IMO having an exclamation mark here is somewhat misleading, as not having WordPress installed is not a strange or unexpected condition.

axmmisaka: Check the PR down below, luckily you didn't make a PR also lool

axmmisaka:
As this issue is fixed by #154 I will close this PR.

PR #141, opened 2020-03-15
check: use CalLinkOrganizations endpoint

ja5087:
I'm not sure why this pointed here since in the past it would go to the help page for CalLinkOrganizations, but the query parameters would be thrown away.


ja5087:
This API doesn't work outside the Berkeley subnet it seems

PR #146, opened 2020-04-24
ocf-kubernetes-deploy: create k8s secrets from file templates

### LINKED ISSUE #145: ocf-kubernetes-deploy: allow for templating large config files with secrets
Extend ocf-kubernetes-deploy so that it will automatically create kubernetes secret resources for certain template files using the same secrets system as other resources. This is needed for config.js in thelounge (instead of ocf/thelounge#9) and homeserver.yaml in matrix.

cg505:
Fixes #145.

This will look for files like `kubernetes/secrets/<secret_name>/*.erb`, and create corresponding Kubernetes secrets containing these files as key-value pairs. This allows us to template files to include injected secrets and then mount the Kubernetes secret resource as a volume into a container.

This allows files like homeserver.yaml in matrix to be stored in the git repo for matrix, and just have the secret portions injected (given the correct volume setup in the deployment yaml).

### DIFF
--- a/staff/sys/ocf-kubernetes-deploy
+++ b/staff/sys/ocf-kubernetes-deploy
@@ -9,8 +9,25 @@ import os
 import subprocess
 import sys
 import tempfile
+import textwrap
 
 import yaml
+
+
+SECRET_RESOURCE_TEMPLATE = '''
+apiVersion: v1
+kind: Secret
+metadata:
+  name: {name}
+type: Opaque
+stringData:
+{string_data}
+'''
+
+STRINGDATA_SECRET_TEMPLATE = '''
+{name}: |2
+{value}
+'''
 
 
 def get_current_context():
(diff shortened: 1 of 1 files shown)


staff/sys/ocf-kubernetes-deploy
```diff
@@ -9,10 +9,27 @@ import os
...
+  name: {name}
+type: Opaque
+stringData:
+{string_data}
+'''
+
+STRINGDATA_SECRET_TEMPLATE = '''
+{name}: |
+{value}
+'''
```
dkess, line 30: Instead of templating yaml, you should build up a json object with standard Python types (dicts, lists, strings) and then convert it to json at the end. All json is valid yaml anyways so there's no need to touch yaml here.

cg505: The problem is that we aren't creating a yaml file, we are creating an erb file. That is to say, we are NOT doing json or yaml manipulation; fundamentally we are doing string manipulation to create an erb file that will be templated into a yaml file. Working in json creates an additional level of conversion that we have to make sure not to break.

A concrete example of what this means:

Say we decide to use a [partial](https://github.com/shopify/krane#using-partials) in our secret template. `ocf-kubernetes-deploy` reads our template and stuff it into a json object. Then we dump the json to a file. Now, the `"` characters in the partial tag in the template have been escaped by the json parser, and the generated ocf-kubernetes-deploy-secret-xyz.json.erb is not valid erb (even though the secret template _is_ valid erb).

This reasoning is definitely a little unintuitive, but it is principled. That said, the template thing I used is a little messy and it would be great if there was a better way to do this.

PR #148, opened 2020-05-03
backup-user: Update NFS host warning, not on pandemic now

jvperrin:
We used to have NFS on pandemic, but that is no longer the case as of quite a few years. Let's just warn when running a user backup on `supernova` that it would be faster on the NFS host but not hardcode a hostname here.

### DIFF
--- a/staff/acct/backup-user
+++ b/staff/acct/backup-user
@@ -1,17 +1,12 @@
 #!/bin/bash -e
 
 [ "$#" -eq 1 ]  || (echo "usage: $0 [account]" >&2; exit 1)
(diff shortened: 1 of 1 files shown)


jvperrin:
Anything with NFS (homes + services) should work fine, but the NFS host itself would be fastest

cg505:
I feel like it's ok to put dataloss, since that's relatively permanent. In any case, let's add death to the list of hosts to print the warning, because it's pretty common to be ssh'd in there debugging web stuff.

jvperrin:
Alright, I've updated this to just check if it's not being run on `dataloss` and warn if so. Otherwise the checks for web/home dirs should catch if it's on a non-NFS host.

Issue #103, opened 2019-04-01
makemysql: automatically update db password in wordpress


dkess:
If a user has a Wordpress website set up and resets their database password, the site becomes completely unusable (even to go into settings and change it). Fixing it requires SSHing in and changing the password in the configuration file.

Luckily, we can automate this with wpcli. The `makemysql` script should also cd into the `public_html` folder and run `wp config set DB_PASSWORD newpasswordgoeshere`. It's OK if the command fails since that just means the user doesn't have a Wordpress site.

kpengboy:
One must take into account security considerations too. Specifically, last time I checked, wp-cli operates by evaluating the Wordpress PHP code, which is dangerous if the Wordpress files have been compromised. We do not want `makemysql` to accidentally run malicious code.

dkess:
Good point. If the site is compromised, it is already able to run malicious code, so we just have to make sure that `wpcli` is run with the same permissions as the user.

kpengboy:
Still, though, I think people expect `makemysql` to be a "safe" command. Allowing it to execute arbitrary code, even as the same user, would still be a major change in the security model of this script.

dkess:
I don't see it as a change in security model at all. There are two cases:

* The site is a regular WordPress site, in which case it works properly
* The site is compromised, in which case it can run whatever code it wants, so interaction with this script doesn't matter

kpengboy:
Still, I suppose there could, at least in principle, be an issue with `makemysql` _triggering_ the execution of malicious code that could, for instance, detect when the database password is being changed through wp-cli and email the new password off to the attacker. Or something.

I would suggest putting the wp-cli stuff behind a command line option, or at least having an option to skip it.

gundralaa:
Hi! Me and snarain worked on a small commit to the initial makemysql script that creates a db pass and assigns it to a temp file that the makemysql-real python script reads. We cant run the modified python script due to privileges but can someone could look at our commit on?
[https://github.com/ocf/utils/tree/wp-automatic-db-pass](url)

dkess:
Thanks for working on this! You should still be able to test the script locally by installing Wordpress on your personal OCF account and running the script from your home directory.

cg505:
Done in #129, although we may want to polish some ends (like not failing when there's no wp installation)

PR #154, opened 2020-05-11
rewrote makemysql so that easywp might work

axmmisaka:
https://github.com/ocf/utils/issues/137
Problem is it used to just output `sudo -u mysql /opt/share/utils/makeservices/makemysql-real`, and after https://github.com/ocf/utils/pull/129 it's nothing, https://github.com/ocf/utils/pull/138 sometimes give a string so this will screw up, as easywp rely on some stuff extracted from `tail -n 1` that look like `Your MySQL database password is: aioeuwhgviujashfuiasb`
This one just simulate if we were running `sudo -u mysql /opt/share/utils/makeservices/makemysql-real` and output the stuff in the last line so there are probably backwards compatibility

another feature: use argument --ignore-wp to not reset in case you don't want to, and this shall remain a secret so that it is not abused

potential problems:
1. Does `(wp core is-installed --quiet --path=$HOME/public_html > /dev/null 2>&1` return true if wp exists and false otherwise
2. Change password if installed with easywp is correct but change password only if installed with easywp is a wrong statement, idk if wp-cli will also handle this
3. idk if it works

### DIFF
--- a/makeservices/makemysql
+++ b/makeservices/makemysql
@@ -6,6 +6,16 @@ fi
 
 PASS=$(sudo -u mysql /opt/share/utils/makeservices/makemysql-real | tee /dev/tty | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
 
-echo 'Changing WordPress database password'
-cd ~/public_html/
-wp config set DB_PASSWORD "$PASS" > /dev/null
+# Check if wp is installed with wp-cli
+# And change password if installed with easywp
+# But change password only if installed with easywp is a wrong statement 
+if $(wp core is-installed --quiet --path=$HOME/public_html > /dev/null 2>&1) &&
+        ([ $# -ne 1 ] || ( [ $# -eq 1 ] && [ ! "$1" = '--ignore-wp' ])); then
+    wp config set DB_PASSWORD "$PASS" > /dev/null 2>&1
+fi
+# Added for backward compatibility - the last line will be
+# extracted by "tail -n 1" and the mysterious "grep"
+# will extract the password after the colon
+# So that it simulate the same behaviour as the original makemysql
+# which output the same thing as in makemysql-real
+echo "Your MySQL database password is: "$PASS


axmmisaka:
Remember to close https://github.com/ocf/utils/issues/103 also
Now CS170 time for me, damn

axmmisaka:
lets see if not running `make test` cause problems again

makeservices/makemysql
```diff
@@ -6,6 +6,18 @@ fi
...
+# And change password if installed with easywp
+# But change password only if installed with easywp is a wrong statement
+# Use --ignore-wp flag to skip this process
+if (wp core is-installed --quiet --path="$HOME/public_html" > /dev/null 2>&1) &&
+        ([ $# -ne 1 ] || ( [ $# -eq 1 ] && [ ! "$1" = '--ignore-wp' ])); then
+    echo "WordPress installation detected, changing WordPress mysql password for you."
+    wp config set DB_PASSWORD "$PASS" > /dev/null 2>&1
+fi
+# Added for backward compatibility - the last line will be
+# extracted by "tail -n 1" and the mysterious "grep"
```
kpengboy, line 19: You may wish to mention which script is running these commands.

makeservices/makemysql
```diff
@@ -6,6 +6,18 @@ fi
...
+# will extract the password after the colon
+# So that it simulate the same behaviour as the original makemysql
+# which output the same thing as in makemysql-real
```
kpengboy, line 22:
```
# So that it simulate the same behaviour as the original makemysql
# which output the same thing as in makemysql-real
```
I would remove these comments; the comments in the previous lines about easywp already tell me everything I need to know.

cg505:
I feel like it would make more sense to update the whole stack here in one swipe:

1. `makemysql-real` mostly prints to stderr, and only prints the password to stdout, which means we don't have to do the scary grep incantation that's currently in `makemysql`
2. `makemysql` has an option to skip WP config, and also an option to _only_ output the password (instead of the normal, nice message "Your MySQL database password is: ")
3. `easywp` uses the options in #2 to get the DB password without doing any weird text manipulation.

axmmisaka:
will do

axmmisaka:
>     3. `easywp` uses the options in #2 to get the DB password without doing any weird text manipulation.

Do u mean 2. here or PR #2...?

cg505:
> Do u mean 2. here or PR #2...?
point 2. here, not the PR. it just autolinked.

makeservices/easywp
```diff
@@ -57,7 +57,7 @@ fi
 
 # Get SQL password
 echo "Resetting database password..."
-sqlpass=$(echo "yes" | makemysql | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
+sqlpass=$(makemysql --unreadable-output --silent)
```
cg505, line 60: seems like you should have `--ignore-wp` here too, since there won't be any wordpress installation yet?

makeservices/makemysql
```diff
@@ -4,20 +4,54 @@ if [[ "$(hostname)" != "tsunami" && "$(hostname)" != "dev-tsunami" ]]; then
...
+            ;;
+        -u|--unreadable-output)
+            UNREADABLE_OUTPUT=true
+            shift
+            ;;
+        -s|--silent)
+            YES="--silent"
+            shift
+            ;;
+        # I've seen -*|--*=), but idk what does the = serve
```
cg505, line 26: it's for options like --key=value probably, but i wouldn't worry about it

makeservices/makemysql
```diff
@@ -4,20 +4,54 @@ if [[ "$(hostname)" != "tsunami" && "$(hostname)" != "dev-tsunami" ]]; then
...
+        -*|--*)
+            # Output to stderr
+            echo "Error: Unsupported flag $1" >&2
+            exit 1
+            ;;
+    esac
+done
+
+PASS=$(sudo -u mysql /opt/share/utils/makeservices/makemysql-real $SILENT | tail -n 1)
```
cg505, line 35: I'm not sure what's the point of --silent, since we're always only going to see the last line (because of tail). Maybe we could make it so that makemysql-real prints its messages to stderr so they aren't affected but the subshell and tail?

makeservices/makemysql
```diff
@@ -4,20 +4,54 @@ if [[ "$(hostname)" != "tsunami" && "$(hostname)" != "dev-tsunami" ]]; then
...
 
 # Check if wp is installed with wp-cli
 # And change password if installed with easywp
 # But change password only if installed with easywp is a wrong statement
 # Use --ignore-wp flag to skip this process
-if (wp core is-installed --quiet --path="$HOME/public_html" > /dev/null 2>&1) &&
-        ([ $# -ne 1 ] || ( [ $# -eq 1 ] && [ ! "$1" = '--ignore-wp' ])); then
-    echo "WordPress installation detected, changing WordPress mysql password for you."
+if (wp core is-installed --quiet --path="$HOME/public_html" > /dev/null 2>&1) && ! $IGNORE_WP ; then
```
cg505, line 41: this is kind of weird, it is literally running the commands `true` or `false`

makeservices/makemysql
axmmisaka: That is what I was expecting (only printing the password), but I do not know if other calls in `makemysql-real` print anything to `stdout`, so `tail` was added to make sure.
That being said, I was considering about what if `makemysql-real` failed, without `grep` `PASS` would be some weird stuffs

makeservices/easywp
axmmisaka: `--ignore-wp` is for those who installed WordPress but don't want to change their pw, here the thing will be done by `wp-cli` in `makemysql` imo

makeservices/makemysql
axmmisaka: lemme do some research, my bash sucks

kpengboy: - I'd check `IGNORE_WP` first before running a command.
 - Nit: beware that putting parentheses around a command runs it in a subshell (which is not necessary in this case). Up to you if you want to keep them around or not, though; I don't think it makes too much of a practical difference in this case either.
 - https://stackoverflow.com/questions/2953646/how-can-i-declare-and-use-boolean-variables-in-a-shell-script

makeservices/makemysql-real
```diff
@@ -78,43 +80,50 @@ def main():
         # Read config file.
         mysql_host, mysql_root_pw = read_config()
 
+        # Added a simple and stupid argument parsing so that other scripts can use it without tunneling in yes.
+        if len(sys.argv) > 1 and sys.argv[1] in ['-s', '--silent']:
```
kpengboy, line 84: Personally, I'd use `argparse`.

makeservices/makemysql-real
```diff
@@ -78,43 +80,50 @@ def main():
...
                                          passwd=mysql_root_pw)
         except MySQLdb.MySQLError:
-            print('>>> Error: Failed to connect to MySQL server.')
-            raise
+            raise ConnectionError('>>> Error: Failed to connect to MySQL server.')
 
         # Check if the database already exists.
         try:
-            print(">>> Checking if database '{}' already exists...".format(username))
+            if not silent:
```
kpengboy, line 103:
```
if not silent:
    print(
```

Probably worth creating a wrapper for this...

makeservices/easywp
kpengboy, line 60: Maybe `--machine-readable` or similar?

makeservices/makemysql-real
axmmisaka: There's only one possible argument which I don't want to utilize a heavy parser

makeservices/makemysql
axmmisaka: I've actually done that, but I didn't test `makemysql-real` and the things it call might output some other mysterious stuff ig

axmmisaka:
Idk what happened to my local repo, but there seems to be a screwup

Updates: 
1. Found a stupid logical mistake I made in `makemysql`. `--machine-readable` does not make sense as it's doing the same thing as `--silent` does, and we always want to `makemysql-real` silently when calling it in the script - and even if we did not run it silently, it's in a subshell so nothing will be outputted... 
2. Added some exit code check after sub-shelling
3. Applied cg505 and kpengboy's suggestions on making script more efficient

Can't believe I procrastinated on this for over 1 year!

makeservices/makemysql
```diff
@@ -4,8 +4,56 @@ if [[ "$(hostname)" != "tsunami" && "$(hostname)" != "dev-tsunami" ]]; then
...
+fi
+
+if $SILENT; then
+    echo "$PASS"
+else
+    # Added for backward compatibility - the last line will be
+    # extracted by "tail -n 1" and the mysterious "grep"
+    # will extract the password after the colon in legacy codes
+    # So that it simulate the same behaviour as the original makemysql
+    # which output the same thing as in makemysql-real
```
kpengboy, line 57: I think this is out of date?

makeservices/makemysql
```diff
@@ -32,20 +26,28 @@ while (( "$#" )); do
...
+    echo 'makemysql did not exit properly. This script will stop.'
+    exit 1
+fi
 
 # Check if wp is installed with wp-cli
 # And change password if installed with easywp
 # But change password only if installed with easywp is a wrong statement
 # Use --ignore-wp flag to skip this process
-if (wp core is-installed --quiet --path="$HOME/public_html" > /dev/null 2>&1) && ! $IGNORE_WP ; then
+if $IGNORE_WP && ! wp core is-installed --quiet --path="$HOME/public_html" > /dev/null 2>&1 ; then
```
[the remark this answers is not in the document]
axmmisaka: Yes, I was stupid!

axmmisaka:
oh no, i decided to rebase and force-push, seems like a very bad decision

Anyway, @kpengboy 's valuable suggestions are applied, and the only thing remaining is pre-commit and style check... Which I need to do on `supernova`

makeservices/easywp
```diff
@@ -57,8 +57,9 @@ fi
...
 # Get SQL password
 echo "Resetting database password..."
-sqlpass=$(echo "yes" | makemysql | tail -n 1 | grep -Po '(?<=: )([0-9a-zA-Z]){24,}$')
-if [[ -z "$sqlpass" ]]; then
+sqlpass=$(makemysql --silent)
+# I don't know if this will work... Just make sure no abnormal exit code first!
+if [ $? -ne 0 ] || [[ -z "$sqlpass" ]]; then
     echo -e "\033[31mError:\033[00m Could not retrieve database password. Run makemysql to see the issue."
     exit 1
 fi
```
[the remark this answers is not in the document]
axmmisaka: Done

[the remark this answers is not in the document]
axmmisaka: Done

makeservices/makemysql-real
```diff
@@ -78,43 +83,45 @@ def main():
         # Read config file.
         mysql_host, mysql_root_pw = read_config()
 
+        # Added a simple and stupid argument parsing so that other scripts can use it without tunneling in yes.
+        if len(sys.argv) > 1 and sys.argv[1] in ['-s', '--silent']:
```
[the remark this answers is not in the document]
axmmisaka: Changed names

ethanhs:
Would you be interested in reviving this PR?

axmmisaka:
Yes, I think it's good tho?

ethanhs:
It appears CI is failing, could you look into that? Also wasn't sure if there were any review comments left to do

kpengboy:
It seems that you'll need to fix some shellcheck issues? Otherwise LGTM, assuming that you have tested this.

axmmisaka:
Remember to SQUASH if you reviewed and decided to merge please 😴

kpengboy (approved):
LGTM, assuming you have tested these changes thoroughly.

PR #155, opened 2020-05-15
add hibp breach sorry message

nikhiljha:
One part of https://github.com/ocf/projects/issues/39, accounts can be automatically sorried if their password is bad.

### DIFF
--- a/staff/acct/sorry/hibp
+++ b/staff/acct/sorry/hibp
@@ -1,0 +1,2 @@
+Your account has been deactivated because the password associated with it
+was reused from somewhere else. Visit ocf.io/s/pwned for more information.


staff/acct/sorry/hibp
```diff
@@ -0,0 +1,2 @@
+Your account has been deactivated because the password associated with it
+was reused from somewhere else. Visit ocf.io/s/pwned for more information.
```
ja5087, line 2: Maybe let them know that it's due to their email address? Not sure if it's obvious

ja5087: I'd also suggest using a non /s/ shorturl since we expect those to be a little bit more permanent

nikhiljha: @ja5087 Nothing to do with their email address, it checks (a part of) the hash of their password against a database. See this [article](https://blog.cloudflare.com/validating-leaked-passwords-with-k-anonymity/) for why that's secure.

ja5087: Ah, I think I misunderstood HIBP then. I thought it was more we'd see if their password + email combo has been pwned, but if it's just a really big rainbow table we should let them know. Either way it's not clear what the requirement is.

nikhiljha: Yeah, I need to write up a page somewhere that explains how it works and why they were locked out.

a big rainbow table is good because if the password was used by someone else, it's definitely not a secure password :)

cg505:
I think it might make more sense to just use have a more general "your password has been leaked" sorryfile, that we can use in other instances when a user's password has been compromised. The downside of that would be that they might think that the password was leaked specifically as their OCF account password, and I don't know if that's important. (Thoughts?)

nikhiljha:
Yeah I'm trying to both...

- not cause unnecessary alarm/panic
- make sure it's clear that the OCF wasn't hacked

staff/acct/sorry/hibp
cg505: agree about `/s/` btw

PR #161, opened 2020-08-02
Fix tests with shellcheck on buster

jvperrin:
I tested `check`, `add-to-group`, and `how`, but have not tested the other scripts apart from testing replacements to make sure they behave the same (`command -v <blah>` vs `which <blah>` for instance, or making sure that echo behaves the same with the extra backslash escaping).


ja5087:
I briefly tested echo with double escaping and it seemed to work the same to me (in the process writing #164). The fact that we arrived at the same PR means shellcheck is pretty good!

ja5087 (approved):

PR #164, opened 2020-10-03
Fixing shell syntax for new shellcheck

ja5087:
Looks like new shellcheck on buster has new rules which are admittedly good style.


jvperrin (approved):
I think I've got essentially the same review in https://github.com/ocf/utils/pull/161 but didn't do much testing with it, so applying either one sounds fine to me

ja5087:
lol yikes I should have checked, thanks jvp. I'll close this.

PR #166, opened 2020-11-26
Add kubernetes TLS cert generation script

fydai:
This adds the required certificate generation script for https://github.com/ocf/puppet/pull/1053.


ethanhs:
I think this will no longer be needed with new kubernetes, right?

fydai:
This is the certificate generation script for the new kubernetes, it is needed.

Issue #168, opened 2021-05-19
Have easywp install a plugin to block REST access


ethanhs:
We've been getting emails about this from campus, and in the long run it will be easier to ask everyone to have it installed.

I propose we install https://wordpress.org/plugins/disable-wp-rest-api/ or https://wordpress.org/plugins/rest-api-toolbox/

after setting up the wordpress install.

We should document this happens on ocfweb and probably say something about it when easywp is being run.

axmmisaka:
If we add plugin-installation to `easywp`, we would have to do configuration for them as well (see below), because otherwise wordpress is not fully installed (they have to access their site to finish installation procedures like setting up an admin account). It'd be weird if we default an admin username/password for them, but otherwise it would be `hardwp`.

Basically, after the current `easywp`, we need to do `wp core install --admin_user=kagamiharan --admin_password=shima_rin_prpr --admin_email=kagamihara@ocf.Berkeley.EDU --skip-email --title="YuruCamp!"` and `wp language core activate zh_CN` if needed.   
Then, we can `wp plugin install disable-json-api --activate`. I prefer this as this gives people a chance to change settings and it is very widely used (70000+ installations iirc), but this time the issue is, as I asked, do we **only** disable unauthenticated REST API access? If no, we would have to write to SQL and change plugin settings - not a hard job, but require some work. Settings are in `wp_options`, with a column that has `option name` `disable_rest_api_options`.  
Finally, this does not prevent users from installing random plugins uses their own REST APIs (like, without using wp core REST API API), not sure if campus security unit wants to permit that - if not, the plugin is not satisfactory.

PR #169, opened 2021-05-27
disable wp REST API in easywp

axmmisaka:
I used this (https://wordpress.org/plugins/disable-json-api/) but not sure how to modify its settings through SQL if such operation is needed...
Also, this looks like `hardwp`, not very `easywp`

https://github.com/ocf/utils/issues/168


makeservices/easywp
```diff
@@ -69,4 +69,30 @@ wp core download
...
...t;user_id&gt;(?:[\d]+|me))/application-passwords";b:0;s:77:"/wp/v2/users/(?P&lt;user_id&gt;(?:[\d]+|me))/application-passwords/introspect";b:0;s:90:"/wp/v2/users/(?P&lt;user_id&gt;(?:[\d]+|me))/application-passwords/(?P&lt;uuid&gt;[\w\-]+)";b:0;s:15:"/wp/v2/comments";b:0;s:35:"/wp/v2/comments/(?P&lt;id&gt;[\d]+)";b:0;s:13:"/wp/v2/search";b:0;s:59:"/wp/v2/block-renderer/(?P&lt;name&gt;[a-z0-9-]+/[a-z0-9-]+)";b:0;s:18:"/wp/v2/block-types";b:0;s:54:"/wp/v2/block-types/(?P&lt;namespace&gt;[a-zA-Z0-9_-]+)";b:0;s:85:"/wp/v2/block-types/(?P&lt;namespace&gt;[a-zA-Z0-9_-]+)/(?P&lt;name&gt;[a-zA-Z0-9_-]+)";b:0;s:15:"/wp/v2/settings";b:0;s:13:"/wp/v2/themes";b:0;s:42:"/wp/v2/themes/(?P&lt;stylesheet&gt;[\w-]+)";b:0;s:14:"/wp/v2/plugins";b:0;s:54:"/wp/v2/plugins/(?P&lt;plugin&gt;[^.\/]+(?:\/[^.\/]+)?)";b:0;s:29:"/wp/v2/block-directory/search";b:0;s:18:"/wp-site-health/v1";b:0;s:43:"/wp-site-health/v1/tests/background-updates";b:0;s:42:"/wp-site-health/v1/tests/loopback-requests";b:0;s:37:"/wp-site-health/v1/tests/https-status";b:0;s:45:"/wp-site-health/v1/tests/dotorg-communication";b:0;s:45:"/wp-site-health/v1/tests/authorization-header";b:0;s:34:"/wp-site-health/v1/directory-sizes";b:0;}}}}'
```
ethanhs, line 97: This seems pretty much unmaintainable. Could you please leave a comment about what this does?

Also is it at all possible to put these settings in some file somewhere?

axmmisaka: Absolutely, this is very distant but I still recall...
This is the settings for `disable-json-api` to work with its maximum potential.
To get it: first, I install `disable-json-api`, then I change its settings so that they fit our needs. I then check what is in `wp_options.disable_rest_api_options`, where settings are hopefully stored. I copy it out then paste it here.

ethanhs: Yeah perhaps reading from a file into the variable would be a bit cleaner. But I guess the settings are just in an unfortunate format.

axmmisaka: the real issue is this does not work...

ethanhs: Ah, yeah that is a problem...

axmmisaka: BTW I don't quite recall what's the context, maybe check here https://github.com/ocf/utils/issues/168 and see if configuring for users are necessary (I remember this only adds extra functions, without configuring there's still some REST APIs disabled)

ethanhs: Yeah the main issue was that security contacted us and informed us that the REST API by default allows user enumeration so we had to tell a couple of vhosts to install the plugin. The hope is that if we install it by default we won't need to deal with a bunch of security emails.

ethanhs: Oh and I think the issue with authenticated vs unauthenticated access was we were worried that it would break plugins if we installed it.

Perhaps for new installations that isn't an issue if we tell people they can allow access for authenticated requests?

axmmisaka:
Squashing and merging, this also (hopefully) makes `easywp` easier as people don't need to manually configure it

PR #170, opened 2021-10-03
Use card by index for ocf-tv volume control

ethanwu10:
The name of the sink changes depending on what profile is selected, so use the sink by index since there's only one sink (right now).


ethanwu10:
Turns out the sink number isn't stable either :(
```
etw@tornado:~$ pactl list sinks
Sink #2
	State: SUSPENDED
	Name: alsa_output.pci-0000_00_03.0.hdmi-stereo
```

Issue #101, opened 2019-03-15
Command for managing printers in rotation


dkess:
We currently have an [annoying manual process](https://www.ocf.berkeley.edu/docs/staff/procedures/printing/#h2_removing-printers-from-service) for taking printers in and out of rotation that's prone to error. We should have a script that does this properly.

Boomaa23:
Can this be done with: `lpadmin -p printername -r classname`? [Source](https://www.cups.org/doc/admin.html#CLASSES)

dkess:
Nice find! Try it, if it works it works :)

Boomaa23:
Tested! It does work! Just to note it's `logjam-double` or `logjam-single` (i.e. per PPD) and use the `-c` flag instead of `-r` to re-add it

PR #174, opened 2022-02-16
Printer rotation command

Boomaa23:
Fix for #101 with regards to https://github.com/ocf/ocfweb/pull/754.

### DIFF
--- a/staff/lab/mod-printer
+++ b/staff/lab/mod-printer
@@ -1,6 +1,7 @@
 #!/usr/bin/env python3
 """Remove or add printers to/from CUPS classes."""
 import argparse
+import re
 import subprocess
 import sys
 
@@ -8,44 +9,50 @@ import sys
 def modify_printer(args):
     """Perform internal call to CUPS CLI to modify class(es)."""
     # Determine which CUPS classes are available
-    cups_classes = subprocess.check_output(["lpstat", "-p"]).decode("utf-8").split("\n")
-    # Remove empty trailing newline
-    cups_classes.remove("")
-    for i in range(len(cups_classes)):
-        clazz = cups_classes[i]
-        # Output is of the form: printer <printername> is idle. <etc>
-        clazz = clazz[len("printer "):]
-        clazz = clazz[:clazz.index(" is ")]
-        cups_classes[i] = clazz
+    try:
+        lpstat_out = subprocess.run(["lpstat", "-p"], check=True, \
+            stdout=subprocess.PIPE, stdin=subprocess.PIPE).strip().split("\n")
+    except subprocess.CalledProcessError as e:
+        print("ERROR: lpstat process exited with exit code %i" % e.returncode)
+        return e.returncode
+    
+    # Parse lpstat output for printer class names
+    # Output is of the form: printer <printername> is idle. <etc>
+    cups_classes = [re.search(r"printer (\S*) is", classname).group(1) for classname in lpstat_out]
 
     if args.verbose:
-        print("Found CUPS classes: " + str(cups_classes))
+        print("Found CUPS classes: %s" % cups_classes)
     
-    args_clazz = getattr(args, "class")
-    if args_clazz and args_clazz in cups_classes:
-        cups_classes = [args_clazz]
-        if args.verbose:
-            print("Overriding CUPS classes with parameter: " + str(args_clazz))
+    # Override with specified classname if provided and valid
+    if args.classname:
+        if args.classname in cups_classes:
+            cups_classes = [args.classname]
+            if args.verbose:
+                print("Overriding CUPS classes with parameter: %s" % args.classname)
+        else:
+            print("ERROR: Specified class %s could not be found in available classes %s" % (args.classname, cups_classes))
+            return 1
     
     # Perform pre-call logging and input processing
     printer = args.printer.lower()
     if args.verbose:
-        print("Printer to perform action on: " + printer)
+        print("Printer to perform action on: %s" % printer)
     action_flag = "-r" if args.action == "remove" else "-c"
     if args.verbose:
-        print("Action to be performed (and flag): " + args.action, "(" + action_flag + ")")
+        print("Action to be performed (and flag): %s (%s)" % (args.action, action_flag))
         
     # Call lpadmin command to perform addition/removal
-    for clazz in cups_classes:
-        action_cmd = subprocess.Popen(["lpadmin", "-p", printer + "-" + clazz, action_flag, clazz])
-        out, err = action_cmd.communicate()
-        if action_cmd.returncode:
-            print("ERROR: lpadmin returned a non-zero value " + str(action_cmd.returncode))
-            return 1
-        if err:
-            print("ERROR: an error occurred while running lpadmin - " + str(err))
-            return 1
-    print("Printer " + printer + " was successfully " + args.action + "(ed) to/from the CUPS class(es): " + str(cups_classes))
+    for classname in cups_classes:
+        action_cmd = ["lpadmin", "-p", printer + "-" + classname, action_flag, classname]
+        try:
+            subprocess.run(action_cmd, check=True, stdout=subprocess.PIPE, stdin=subprocess.PIPE)
+        except subprocess.CalledProcessError as e:
+            print("ERROR: lpstat process exited with exit code %i" % e.returncode)
+            return e.returncode
+        else:
+            action_out = "added to" if args.action == "add" else "removed from"
+            print("Printer %s was successfully %s %s" % (printer, action_out, classname))
+            
 
 def main():
     args = argparse.ArgumentParser(description=__doc__)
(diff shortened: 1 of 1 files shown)


staff/lab/mod-printer
```diff
@@ -0,0 +1,69 @@
...
+"""Remove or add printers to/from CUPS classes."""
+import argparse
+import subprocess
+import sys
+
+
+def modify_printer(args):
+    """Perform internal call to CUPS CLI to modify class(es)."""
+    # Determine which CUPS classes are available
+    cups_classes = subprocess.check_output(["lpstat", "-p"]).decode("utf-8").split("\n")
```
dkess, line 11: Use the more modern `subprocess.run` function here. It'll end up being cleaner since it'll automatically decode the output, you can configure it to throw an exception if there's an error exit code, etc.

staff/lab/mod-printer
```diff
@@ -0,0 +1,69 @@
...
+    # Remove empty trailing newline
+    cups_classes.remove("")
```
dkess, line 13: Better way:

```py
subprocess.run(...).strip().split('\n')
```
In other words: strip the whitespace before splitting, then you don't have to delete it on a second line.

staff/lab/mod-printer
```diff
@@ -0,0 +1,69 @@
...
+    for i in range(len(cups_classes)):
+        clazz = cups_classes[i]
+        # Output is of the form: printer <printername> is idle. <etc>
+        clazz = clazz[len("printer "):]
+        clazz = clazz[:clazz.index(" is ")]
+        cups_classes[i] = clazz
```
dkess, line 19: This makes the code harder to read, since you're reusing the `cups_classes` variable to have two purposes and two different meanings.

I'd recommend making the first variable called `command_output` or something like that, and the second variable can be `cups_classes` (the actual list of class names).

Also recommend using a regex to parse the command output lines instead of this kind of string manipulation.

staff/lab/mod-printer
```diff
@@ -0,0 +1,69 @@
...
+
+def main():
+    args = argparse.ArgumentParser(description=__doc__)
+    args.add_argument("action",
+        choices=["add", "remove"], 
+        help="Action to perform"
+    )
+    args.add_argument("printer", help="Name of printer to modify")
+    
+    args.add_argument("-c", "--class", help="Remove from only a specific class")
```
dkess, line 58: Rename this to `classname` (or something else) so you don't have to retrieve it with the awkward `getattr(args, "class")` thing.

staff/lab/mod-printer
```diff
@@ -0,0 +1,69 @@
...
+        # Output is of the form: printer <printername> is idle. <etc>
+        clazz = clazz[len("printer "):]
+        clazz = clazz[:clazz.index(" is ")]
+        cups_classes[i] = clazz
+
+    if args.verbose:
+        print("Found CUPS classes: " + str(cups_classes))
+    
+    args_clazz = getattr(args, "class")
+    if args_clazz and args_clazz in cups_classes:
```
dkess, line 25: If `args_clazz` isn't in the class list, this code will act as if the arg wasn't specified, when in reality it means the user probably made a mistake, which could lead to unintended consequences. Instead, if it's not in the class list, return an error early and tell the user it doesn't exist.

staff/lab/mod-printer
```diff
@@ -0,0 +1,69 @@
...
+    printer = args.printer.lower()
+    if args.verbose:
+        print("Printer to perform action on: " + printer)
+    action_flag = "-r" if args.action == "remove" else "-c"
+    if args.verbose:
+        print("Action to be performed (and flag): " + args.action, "(" + action_flag + ")")
+        
+    # Call lpadmin command to perform addition/removal
+    for clazz in cups_classes:
+        action_cmd = subprocess.Popen(["lpadmin", "-p", printer + "-" + clazz, action_flag, clazz])
```
dkess, line 40: Again, use `subprocess.run` here. You can use the `check=True` option to raise on an error code.

staff/lab/mod-printer
```diff
@@ -0,0 +1,69 @@
...
+        if args.verbose:
+            print("Overriding CUPS classes with parameter: " + str(args_clazz))
+    
+    # Perform pre-call logging and input processing
+    printer = args.printer.lower()
+    if args.verbose:
+        print("Printer to perform action on: " + printer)
+    action_flag = "-r" if args.action == "remove" else "-c"
+    if args.verbose:
+        print("Action to be performed (and flag): " + args.action, "(" + action_flag + ")")
```
dkess, line 36: Use [f-strings](https://docs.python.org/3/tutorial/inputoutput.html#formatted-string-literals), here and throughout.

staff/lab/mod-printer
```diff
@@ -0,0 +1,69 @@
...
+    for clazz in cups_classes:
+        action_cmd = subprocess.Popen(["lpadmin", "-p", printer + "-" + clazz, action_flag, clazz])
+        out, err = action_cmd.communicate()
+        if action_cmd.returncode:
+            print("ERROR: lpadmin returned a non-zero value " + str(action_cmd.returncode))
+            return 1
+        if err:
+            print("ERROR: an error occurred while running lpadmin - " + str(err))
+            return 1
+    print("Printer " + printer + " was successfully " + args.action + "(ed) to/from the CUPS class(es): " + str(cups_classes))
```
dkess, line 48: Nit: conventionally CLI tools are quiet on success (unless verbose mode is on). Up to you if you want to have this success message. But if you do want to have a success message, I'd recommend doing one per for loop. That way if the procedure failed mid-way through, the user can see how it partially succeeded.

dkess (changes requested):
Nice job, thanks for working on this! Have some style nits but this is definitely the right idea.

dkess:
One more thing: it might later be useful to add a subcommand that lists the current printer state (i.e. tell the user which printers are in/out of service). You might want to look into argparse subparsers (https://docs.python.org/3/library/argparse.html#sub-commands) for a better way of handling this (as opposed to a conditional arg for "add"/"remove"

staff/lab/mod-printer
Boomaa23: I normally would use f-strings, but in this case things seem to be running Python 3.5, which does not have support for f-strings. :(

dkess: Does this only work on whiteout and not supernova?

If so then... ugh, yeah, forgot about that. Though I will say, whiteout is probably a relatively easy server to do a buster/bullseye upgrade on.

Boomaa23: Yeah... Maybe in the future I can do a buster/bullseye upgrade on whiteout, but at the moment that might be a bit much.

Boomaa23:
Okay so I took your changes/suggestions and now it uses a subparser, with functionality to list/add/remove. Notably however CUPS doesn't provide a CLI option for listing "non-shared" printers (i.e. hardware printers, not just the classes), so the best I could do was just print the printers that are currently active. Also the entire thing is now more modern, as per your suggestions. Thanks for your help with this @dkess!

staff/lab/mod-printer
```diff
@@ -0,0 +1,131 @@
...
+import sys
+
+
+def modify_printer(args):
+    """Perform internal call to CUPS CLI to modify class(es)."""
+    # Determine which CUPS classes are available
+    try:
+        lpstat_out = subprocess.run(["lpstat", "-p"], check=True, \
+            stdout=subprocess.PIPE, stdin=subprocess.PIPE).strip().split("\n")
+    except subprocess.CalledProcessError as e:
```
dkess, line 15: IMO, it's not necessary to catch the exception. If the command fails, the exception message that gets printed by default will show the exit code, as well as the full argument list.

(same throughout)

staff/lab/mod-printer
```diff
@@ -0,0 +1,131 @@
...
+        print("ERROR: lpstat process exited with exit code %i" % e.returncode)
+        return e.returncode
+
+    # Parse output to fit printer arg, or send directly to stdout
+    status_msgs = lpstat_out.stdout.decode("utf-8")
+    if args.printer:
+        status_lines = status_msgs.split("\n")
+        for line in status_lines[:]:
+            if args.printer not in line and "members of class" not in line:
+                status_lines.remove(line)
```
dkess, line 90: No need to mutate the list, since you aren't using it anywhere else. You can print lines while looping, and skip printing the ones that don't fit the criteria.

staff/lab/mod-printer
```diff
@@ -0,0 +1,131 @@
...
+    lpstat_cmd = ["lpstat", "-c"]
+
+    # Add on to lpstat command with arguments
+    if args.classname:
+        if args.verbose:
+            print("Limiting results to only class %s" % args.classname)
+        lpstat_cmd.append(args.classname)
+    if args.jobs:
+        if args.verbose:
+            print("Including jobs in stdout")
```
dkess, line 69: This print statement (and the one above) is a bit too verbose. Printing the full command (as you do in the next print statement) is sufficient.

dkess:
Re: the build errors, the `make test` will run pre-commit for you and should fix them automatically.

dkess (approved):

PR #178, opened 2022-07-13
Find max OCF UID

Boomaa23:
I had to bump `_KNOWN_UID` ([here](https://github.com/ocf/ocflib/blob/47ff54d753a597972a42cde2e90caf6633178bd2/ocflib/account/creation.py#L27)) recently, and to do so find the maximum currently in use OCF UID from LDAP. Realistically this should be simple enough, but the true maximum value is excluded so more processing is required. This script should automate that process.

### DIFF
--- a/staff/acct/max-uid
+++ b/staff/acct/max-uid
@@ -2,24 +2,34 @@
 """Find the highest non-excluded OCF UID."""
 import sys
 
+from ocflib.account.creation import IGNORED_UID_RANGES
+from ocflib.account.creation import RESERVED_UID_RANGES
 from ocflib.infra.ldap import ldap_ocf
 from ocflib.infra.ldap import OCF_LDAP_PEOPLE
-from ocflib.account.creation import RESERVED_UID_RANGES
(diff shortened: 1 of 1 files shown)


cg505:
Historically you would just look in #atool for the most recently created account, but this is cool too.

staff/acct/max-uid
```diff
@@ -0,0 +1,35 @@
...
+    for start, end in IGNORED_UID_RANGES:
+        if start <= uid <= end:
+            return False
+    return True
+
+
+def main():
+    print('Searching for maximum currently used OCF UID')
+    with ldap_ocf() as c:
+        c.search(
```
axmmisaka, line 24: Similar to what @cg505 said this is fine but indeed a little slow tbh; if we are not running this constantly I guess it's fine?

axmmisaka (changes requested):
Alternatively, since this is same as https://github.com/ocf/ocflib/blame/889145860a908e23610e5b4edd7774691a1beab2/ocflib/account/creation.py#L75, maybe abstract it into a separate function in `ocflib`.

PR #179, opened 2022-09-22
Rewrite pythonic sorry

axmmisaka:
#130


axmmisaka:
This has been 4 years, someone please review it 😥

I'm now curious do we *really* want to rewrite stuff in python?
The main issue is that we use LDAP+Kerberos/GSSAPI which requires us to shell out.

PR #183, opened 2023-08-24
Display group status in `signat`

Kalissaac:


Kalissaac:
Depends on ocf/ocflib#255

PR #188, opened 2025-03-11
Paper update 

laksith19:
Hmm so do we break backwards compatibility or use something other than python argparse?


RonitNath:
use something else(TM)
and also rewrite all of ocf infra again again

xa1on:
updated to retain original syntax. removed default value for view in favor of prompts.

RonitNath (approved):
lgtm. assuming this solves previous issue with argparse, @something please make note of if this isn't the case, otherwise will merge soon

RonitNath:
also thoughts on removing reason as required? has also been part of discussion about this

acct/check
```diff
@@ -166,7 +166,7 @@ fi
 
 # check print quota
 echo
-paper view "$user"
+paper view "-u $user"
```
kpengboy, line 169: Revert?

RonitNath: good catch, assuming should be reverted

xa1on:
> lgtm. assuming this solves previous issue with argparse, @something please make note of if this isn't the case, otherwise will merge soon

what previous issue are you referring to? @RonitNath

acct/check
xa1on: oops. did not notice this. my bad.

xa1on: > good catch, assuming should be reverted

fixed @RonitNath

xa1on (approved):

PR #202, opened 2026-04-24
Update makemysql user check for carp

skybound0:
SUDO_USER does not exist as an accessible environment variable on Nix for the Python os module. Replaced with pwd username check.

### DIFF
--- a/makeservices/makemysql-real
+++ b/makeservices/makemysql-real
@@ -18,6 +18,7 @@ is used to parse the config file, so the format is basically the Windows
 INI format.
 """
 import os
+import pwd
 import random
 import string
 import sys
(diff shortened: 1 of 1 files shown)


jaysa68 (approved):

kpengboy:
Are you sure this works? On Debian at least, `sudo python3 -c "import os; print(os.getuid())"` prints 0.

skybound0:
Yes. Tested and works on tsunami and koi.

sophiebsw:
this does not work properly and is not needed

SUDO_USER is properly populated even on nixos, and the code in question will incorrectly return the uid of the mysql user

PR #205, opened 2026-04-29
Fix makemysql username check

skybound0:
made a mistake using pwd previously, now passes the username of the user initiating the script to makemysql-real


skybound0:
deprecated for ocf/nix#300

PR #207, opened 2026-05-03
update carp/koi references

vqbc:


staff/sys/migrate-vm
jaysa68, line 1: maybe leave the example for this command alone, since im not sure if anyone has actually used it to migrate a VM in some time and im unsure if it would work with our current nixOS VM setup. i dont want to give the false impression that it's been used or updated / checked for functionality since koi was implemented, lol

jaysa68 (approved):
see migrate-vm comment, but then you're good to go

PR #208, opened 2026-05-08
fix makemysql on nix

skybound0:
fix SUDO_USER not existing as environment variable on nix, and incorrect pwd call that returns username 'mysql' due to sudo call in makemysql

fr ong no cap fixed this time


jaysa68 (approved):
bet

kafu-densha:
@jaysa68 is this pr still relevant or did it get fixed in another way?

